PR #4, opened 2018-03-30
Move data.ts to core, cleanup, and fixes.

adarob:


core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@magenta/core",
-  "version": "0.0.9",
+  "version": "0.0.13",
```
cghawthorne: We've kind of being doing versioning and publishing ad hoc, but I propose we move toward:
- Test changes locally using `yarn link` so you don't have to publish a package to use it when developing dependent packages.
- Do PRs without bumping version numbers.
- Once things have been reviewed and checked in, do `yarn publish`, which will generate a commit that just bumps the version number and publishes to npm.
- Define a `prepublish` task for all our packages that runs lint, test, etc.

This aligns us with how we do Python releases and generally makes sure code is reviewed before it's published.

Thoughts?

core/src/index.ts
```diff
@@ -22,5 +22,6 @@ import INote = NoteSequence.INote;
 
 export {NoteSequence, INoteSequence, INote};
 
+export * from './data';
```
cghawthorne: What would you think of:

```
import * as data from './data';
export {data};
```

The `data` package has lots of classes in it, and this puts them under their own namespace. I kind of did this by accident in `midi_io` and `sequences` by putting a bunch of static methods in their own enclosing class. I'm now wondering if it would be better to just have the static methods at the top level and do a similar style of namespace encapsulation as I proposed above.

music_rnn/.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 es5/
 yarn-error.log
+yarn.lock
```
cghawthorne: ditto

music_vae/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules/
+es5/
+yarn-error.log
+yarn.lock
```
cghawthorne: ditto

core/.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 es5/
 yarn-error.log
+yarn.lock
```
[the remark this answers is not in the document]
adarob: Done.

core/package.json
adarob: That sounds good. I wasn't sure how to test that the core changes work in music_rnn, etc. How does yarn link do that?

core/src/index.ts
adarob: Yes, I like this better.

core/package.json
cghawthorne: https://yarnpkg.com/lang/en/docs/cli/link/

In core, you do `yarn link`, which creates a global symlink to make it available to other packages. In music_rnn, you do `yarn link @magenta/core`, which instructs it to use the symlink version, not the installed version. You have to redo this after every time you do `yarn install`, so it's a little annoying, but I think it's better than publishing lots of extra versions.

core/src/data.ts
```diff
@@ -134,19 +134,20 @@ export class DrumsConverter extends DataConverter{
   }
 
   toTensor(noteSequence: INoteSequence) {
-    const drumRoll = dl.buffer([this.numSteps, this.pitchClasses.length + 1]);
+    const numSteps = this.numSteps | noteSequence.totalQuantizedSteps;
```
cghawthorne: did you mean `||`?

music_rnn/src/model.ts
```diff
@@ -15,55 +15,46 @@
...
-import * as data from './data';
-import { INoteSequence } from '@magenta/core';
-
-// tslint:disable-next-line:max-line-length
-const CHECKPOINT_URL = "https://storage.googleapis.com/download.magenta.tensorflow.org/models/music_rnn/dljs/basic_rnn/";
-
-const DEFAULT_MIN_NOTE = 48;
-const DEFAULT_MAX_NOTE = 83;
+import tf = magenta.tf;
+import { isNullOrUndefined } from 'util';
```
cghawthorne: where is util coming from?

adarob: not sure...it must be a dependency somewhere? i found in in autocomplete

core/src/data.ts
adarob: done

music_rnn/src/model.ts
cghawthorne: Ah, looks like it's a node thing. It will work locally, but I think we need to add a dependency on the 'util' package for bundling purposes.

PR #12, opened 2018-04-06
Add TrioConverter, use custom tf split function.

adarob:


core/src/data.ts
```diff
@@ -77,9 +81,10 @@ export function converterFromSpec(spec: ConverterSpec) {
  * `NoteSequence` objects.
  */
 export abstract class DataConverter {
-  abstract numSteps: number;  // Total length of sequences.
-  abstract numSegments: number;  // Number of steps for conductor.
-  abstract readonly NUM_SPLITS: number;  // Number of conductor splits.
+  abstract readonly numSteps: number;  // Total length of sequences.
+  abstract readonly numSegments: number;  // Number of steps for conductor.
+  abstract readonly depth: number;  // Size of final output dimension.
+  abstract readonly NUM_SPLITS: number;  // Const number of conductor splits.
```
cghawthorne: now that the other properties are also readonly, should they either all be uppercase or should NUM_SPLITS be lowercase?

core/src/sequences.ts
```diff
@@ -74,6 +74,10 @@ export class Sequences {
     return n && (n & (n - 1)) === 0;
   }
 
+  public static clone(ns:INoteSequence) {
```
cghawthorne: Thanks, makes sense to centralize this logic. Could you also update sequences_test.ts to use this in the few places it clones NoteSequences?

core/src/tf_lib.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as tf from '@tensorflow/tfjs';
+
+export function split(
```
cghawthorne: is this something we should contribute back to tfjs?

core/src/data.ts
adarob: No, NUM_SPLITS is still something different. It really should be const but const isn't allowed in the abstract definition...

core/src/tf_lib.ts
adarob: Working on it now :)

adarob: https://github.com/tensorflow/tfjs-core/pull/930

adarob:
#9

PR #15, opened 2018-04-08
DrumsOneHotConverter

teropa:
Add a `DrumsOneHotConverter` for the one-hot encoding used by drums-rnn.

Expects `converter.json` for drums-rnn to specify the new one-hot multidrum encoding, like:

```
{
  "type": "DrumsOneHotConverter",
  "args": {
    "pitchClasses": null,
    "numSteps": null
  }
}
````

Related to #10, #14

### DIFF
--- a/core/src/data.ts
+++ b/core/src/data.ts
@@ -220,8 +220,8 @@ export class DrumRollConverter extends DrumsConverter {
  * The `Tensor` output by `toTensor` is a 2D one-hot encoding. Each
  * row is a time step, and each column is a one-hot vector where each drum
  * combination is mapped to a single bit of a binary integer representation,
- * where the bit has value 0 if the drum type is not present, and 1 if it is
- * present.
+ * where the bit has value 0 if the drum combination is not present, and 1 if
+ * it is present.
  *
  * The expected `Tensor` in `toNoteSequence` is the same kind of one-hot
  * encoding as the `Tensor` output by `toTensor`.
@@ -229,13 +229,6 @@ export class DrumRollConverter extends DrumsConverter {
  * The output `NoteSequence` uses quantized time and only the first pitch in
  * pitch class are used.
  *
- * @param numSteps The length of each sequence.
- * @param numSegments (Optional) The number of conductor segments, if
- * applicable.
- * @param pitchClasses (Optional) An array of arrays, grouping together MIDI
- * pitches to treat as the same drum. The first pitch in each class will be used
- * in the `NoteSequence` returned by `toNoteSequence`. A default mapping to 9
- * classes is used if not provided.
  */
 export class DrumsOneHotConverter extends DrumsConverter {
   readonly depth: number;
@@ -277,12 +270,12 @@ export class DrumsOneHotConverter extends DrumsConverter {
       const bin = (labels[s] >>> 0).toString(2);
       for (let i = bin.length - 1; i >= 0; i--) {
         if (bin[i] === '1') {
-          noteSequence.notes.push(
-            NoteSequence.Note.create({
-              pitch: this.pitchClasses[bin.length - i - 1][0],
-              quantizedStartStep: s,
-              quantizedEndStep: s + 1,
-            isDrum: true}));
+          noteSequence.notes.push(NoteSequence.Note.create({
+            pitch: this.pitchClasses[bin.length - i - 1][0],
+            quantizedStartStep: s,
+            quantizedEndStep: s + 1,
+            isDrum: true
+          }));
         }
       }
     }
--- a/core/src/data.ts
+++ b/core/src/data.ts
@@ -70,6 +70,8 @@ export function converterFromSpec(spec: ConverterSpec) {
     return new DrumRollConverter(spec.args as DrumsConverterArgs);
   } else if (spec.type === 'TrioConverter') {
     return new TrioConverter(spec.args as TrioConverterArgs);
+  } else if (spec.type === 'DrumsOneHotConverter') {
+    return new DrumsOneHotConverter(spec.args as DrumsConverterArgs);
   } else {
     throw new Error(
       'Unknown DataConverter type in spec: ' + spec.type);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: music_rnn/src/model.ts [hunk 1]
```diff
@@ -131,41 +155,160 @@ export class MusicRNN {
...
-          // Save samples as bool to reduce data sync time.
-          samples.push(nextInput.as1D().toBool());
-        }
-        [c, h] = tf.multiRNNCell(this.lstmCells, nextInput, c, h);
-      }
-
+      const samples = this.attnMatrix ?
+        this.sampleRnnWithAttention(inputs, steps, temperature) :
+        this.sampleRnn(inputs, steps, temperature);
       return tf.stack(samples).as2D(samples.length, outputSize);
```
teropa: These implementations are different enough that might be appropriate to split them out into separate classes. In Python there's a standalone `AttentionCellWrapper` that manages attention transparently. Using such a thing here may require changing the abstractions a bit.

music_rnn/src/model.ts
```diff
@@ -131,41 +155,160 @@ export class MusicRNN {
...
+    const outputSize:number = inputs.shape[1];
+
+    let c: tf.Tensor2D[] = [];
+    let h: tf.Tensor2D[] = [];
+    for (let i = 0; i < this.biasShapes.length; i++) {
+      c.push(tf.zeros([1, this.biasShapes[i] / 4]));
+      h.push(tf.zeros([1, this.biasShapes[i] / 4]));
+    }
+    const attnSize = this.biasShapes[0] / 4;
+    const attnLength = 32;
```
teropa: Can the attention length be inferred from the model somehow?

music_rnn/src/model.ts
```diff
@@ -131,41 +155,160 @@ export class MusicRNN {
...
+        // Save samples as bool to reduce data sync time.
+        samples.push(nextInput.as1D().toBool());
+      }
+
+      const binaryCounts = tf.buffer([1, numCounterBits]);
+      for (let bit = 0; bit < numCounterBits; bit++) {
+        const counterValue = Math.floor((i + 1) / 2 ** bit) % 2 ? 1.0 : -1.0;
+        binaryCounts.set(counterValue, 0, bit);
+      }
+      nextInput = nextInput.concat(binaryCounts.toTensor(), 1) as tf.Tensor2D;
```
teropa: The binary counter should be in a data converter rather than here IMO, but need to address how to apply it when using the last output as the next input. We don't use the data converters for that.

music_rnn/src/model.ts
adarob: Agreed. Let's split this into two PRs. Just make the drum converter change in this PR for now.

core/src/data.ts
```diff
@@ -210,6 +212,79 @@ export class DrumRollConverter extends DrumsConverter {
   }
 }
 
+/**
+ * Converts between a quantized `NoteSequence` containing a drum sequence
+ * and the `Tensor` objects used by `MusicRNN`.
+ *
+ * The `Tensor` output by `toTensor` is a 2D one-hot encoding. Each
+ * row is a time step, and each column is a one-hot vector where each drum type
```
adarob: s/drum type/drum combination/g

core/src/data.ts
```diff
@@ -210,6 +212,79 @@ export class DrumRollConverter extends DrumsConverter {
...
+    const labelsTensor = oh.argMax(1);
+    const labels: Int32Array = await labelsTensor.data() as Int32Array;
+    labelsTensor.dispose();
+    for (let s = 0; s < labels.length; ++s) {  // step
+      const bin = (labels[s] >>> 0).toString(2);
+      for (let i = bin.length - 1; i >= 0; i--) {
+        if (bin[i] === '1') {
+          noteSequence.notes.push(
+            NoteSequence.Note.create({
+              pitch: this.pitchClasses[bin.length - i - 1][0],
```
adarob: isDrum: true

core/src/data.ts
```diff
@@ -210,6 +212,79 @@ export class DrumRollConverter extends DrumsConverter {
...
+ * applicable.
+ * @param pitchClasses (Optional) An array of arrays, grouping together MIDI
+ * pitches to treat as the same drum. The first pitch in each class will be used
+ * in the `NoteSequence` returned by `toNoteSequence`. A default mapping to 9
+ * classes is used if not provided.
+ */
+export class DrumsOneHotConverter extends DrumsConverter {
+
+  constructor(args: DrumsConverterArgs) {
+    super(args);
```
adarob: We need to set the depth to 2 ** pitchClasses.length.
Since it is read-only, we may just want to not making this extend DrumsConverter and give it its own constructor.

core/src/data.ts
```diff
@@ -210,6 +212,79 @@ export class DrumRollConverter extends DrumsConverter {
...
+  }
+
+  toTensor(noteSequence: INoteSequence) {
+    const numSteps = this.numSteps || noteSequence.totalQuantizedSteps;
+    const indexes = this.toOneHotIndexes(noteSequence);
+    const drumRoll = tf.buffer([numSteps, 2 ** this.pitchClasses.length]);
```
adarob: use a different name than drumRoll

core/src/data.ts
adarob: Actually, I think you can just re-define depth as a readonly property in this class to allow it to be overwritten.

teropa:
Addressed comments and pulled out attention from this branch, another PR to follow with that.

core/src/data.ts
```diff
@@ -211,6 +213,83 @@ export class DrumRollConverter extends DrumsConverter {
...
+ * where the bit has value 0 if the drum type is not present, and 1 if it is
+ * present.
+ *
+ * The expected `Tensor` in `toNoteSequence` is the same kind of one-hot
+ * encoding as the `Tensor` output by `toTensor`.
+ *
+ * The output `NoteSequence` uses quantized time and only the first pitch in
+ * pitch class are used.
+ *
+ * @param numSteps The length of each sequence.
```
adarob: You can leave these params out since they're not exposed here.

core/src/data.ts
```diff
@@ -211,6 +213,83 @@ export class DrumRollConverter extends DrumsConverter {
...
 }
 
+/**
+ * Converts between a quantized `NoteSequence` containing a drum sequence
+ * and the `Tensor` objects used by `MusicRNN`.
+ *
+ * The `Tensor` output by `toTensor` is a 2D one-hot encoding. Each
+ * row is a time step, and each column is a one-hot vector where each drum
+ * combination is mapped to a single bit of a binary integer representation,
+ * where the bit has value 0 if the drum type is not present, and 1 if it is
```
adarob: s/drum type/drum combination/g

adarob:
It looks like there are some formatting issues as well. Can you run `clang-format`? If you are using VS Code, you can just do `yarn install` in /core and open the entire repo as your workspace, and it should do it automatically for you on Save.

teropa:
> t looks like there are some formatting issues as well. Can you run clang-format? If you are using VS Code, you can just do yarn install in /core and open the entire repo as your workspace, and it should do it automatically for you on Save.

Yes.. I'm using Prettier on other projects, will need to figure out how to disable it per project, otherwise it'll format a bunch of stuff on these files 😄 

`clang-format` also turned the exponentiation operator `**` into `* *` for me, which is invalid. Switched them back before committing. If it becomes a problem can just switch to `Math.pow`.

adarob:
Yes, that is really annoying that it does that. I switched to Math.pow elsewhere.

PR #16, opened 2018-04-09
add chord encoders that map chord symbol to 1D tensor

iansimon:


core/src/chords.ts
```diff
@@ -102,3 +111,100 @@ export class ChordSymbols {
...
+    }
+
+    const root = ChordSymbols.root(chord);
+    const quality = ChordSymbols.quality(chord);
+    const index = 1 + quality * constants.NUM_PITCH_CLASSES + root;
+
+    if (index >= this.depth) {
+      throw new ChordEncodingException(
+          'Chord is not a standard triad: ' +
+          `${chord}`);
```
adarob: this can just go inside the above string

core/src/chords.ts
```diff
@@ -102,3 +111,100 @@ export class ChordSymbols {
...
+    }
+
+    const root = ChordSymbols.root(chord);
+    const quality = ChordSymbols.quality(chord);
+    const index = 1 + quality * constants.NUM_PITCH_CLASSES + root;
+
+    if (index >= this.depth) {
+      throw new ChordEncodingException(
+          'Chord is neither major nor minor: ' +
+          `${chord}`);
```
adarob: This can just go inside the above string.

PR #18, opened 2018-04-10
Add AttentionWrapper

### LINKED ISSUE #10: Add support for attention models to MusicRNN.
@teropa, do you want to take this one too or should I?

teropa:
Building on #15 

This version is fully functional with the drums-rnn one-hot checkpoint.

Open questions:
- How to extract the binary counter
- How to parameterize or infer the attention length 

Closes #10


adarob:
@iansimon has a solution for the binary counter that he will work on soon.
Good question on the attention length. Worst case we could add another file to the checkpoint.

iansimon:
I'm planning to support both the binary counter and chord conditioning by allowing the DataConverter to also return an optional time-varying control signal that gets appended to the input.

teropa:
OK using the `BinaryCounter` control signal now. Need wisdom on the following:

1. I'm passing `ControlSignal` as an argument to `continueSequence`, leaving it for the user to construct. Is this how you intended it? For chords this seems perfect to me. For the binary counter maybe less so since it's more a model implementation detail.
2. Attention length also now passed in as an argument to `MusicRNN`, so user will need to know this.
3. Using `tf.softmax` yields more or less random results for me. I'm getting acceptable results by skipping it and passing the logits directly into `tf.multinomial`. Not sure if this is correct as `tf.multinomial` is as of yet undocumented.

adarob:
3. tf.multinomial expects the logits, not the probs. It does its own softmax.

I'll get back to you later today on the other Qs once I have time to think about it more.

adarob:
For 1, I think we should rename ControlSequence to CControlConverter and have it be specified in the constructor or a JSON in the checkpoint like we do for the DataConverter. controlSequence in continueSequence can then become a ControlSequence interface that varies depending on the converter. We would probably want to change the interface to ControlConverter so it just processes the controlSequence all at once like we do with the DataConverter. How does this sound @iansimon?

iansimon:
I think that makes sense, though I'm not sure I would call it a "converter" as it's not converting from the same representation e.g. `NoteSequence` but just depends on a few parameters like `numBits` for the binary counter and the chord progression (and `ChordEncoder`) for chord conditioning.

adarob:
Okay, `ControlEncoder`? 

For the attention length, let's assume it's passed in the constructor or available in the checkpoint dir.
@teropa any other blockers?

teropa:
> For the attention length, let's assume it's passed in the constructor or available in the checkpoint dir.
@teropa any other blockers?

Nope. To have drums-rnn work there will be some things needed in the `converter.json`, and some that the user will need to do.

The checkpoint `converter.json` needs to specify the one-hot encoder, e.g. 

```
{
  "type": "DrumsOneHotConverter",
  "args": {
    "pitchClasses": null,
    "numSteps": null
  }
}
```

The user, when constructing the `MusicRNN`, will need to specify an attention length:

```
const drumsRnn = new MusicRNN(DRUMS_CHECKPOINT, null, 32)
```

As well as use a `BinaryCounter` control with the right number of steps and counters when continuing a sequence:

````
await drumsRnn.continueSequence(qns, 20, null, new magenta.controls.BinaryCounter(qns.notes.length + 20, 6));
```

These are used in the `demo/index.ts` file so once the `converter.json` change is applied on the hosted drums-rnn checkpoint the should be fully functional.

adarob:
I can update the `converter.json` in a moment.

However, I was proposing that we also include files with the checkpoint to specify other attributes that are tied to the train model. In this case we would want a file like `attn_length` with the attention length and a `control.json` file which is similar to `converter.json` in that it specifies the type of control sequence (e.g., "BinaryCounter") and any constructor arguments.

teropa:
Ah yes, SGTM. 

Does the control abstraction allow for user control as well! I.e. for chords I want to supply the control chords when I continue a sequence?

PR #23, opened 2018-04-12
Add basic tests for DataConverters and fix bugs they reveal.

adarob:


core/src/data.ts
```diff
@@ -212,6 +215,7 @@ export class DrumRollConverter extends DrumsConverter {
         }
       }
     }
+    roll.dispose();
```
iansimon: Not that familiar with how this stuff works in Javascript but it seems very strange to dispose a function parameter inside the function.

adarob: indeed! i'd already fixed this but nice catch!

PR #25, opened 2018-04-13
add binary counter and chord progression control signals

iansimon:


core/src/controls.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Binary counter control signal.
+ *
+ * @param numSteps The length of the control signal.
+ * @param numBits The number of binary counter bits.
+ */
+export class BinaryCounter extends ControlSignal {
+  constructor(numSteps: number, numBits: number) { super(numSteps, numBits); }
+
+  tensor(step: number) {
```
adarob: can you add doc here (and below)

core/src/controls.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Abstract ControlSignal class for producing control tensors.
+ *
+ * @param numSteps The length of the control signal.
+ * @param depth The size of the control tensor at each step.
+ */
+export abstract class ControlSignal {
+  readonly numSteps: number;  // Total length of sequences.
+  readonly depth: number;     // Size of final output dimension.
+  abstract tensor(step: number): tf.Tensor1D;
```
adarob: I'd prefer getTensor so this doesn't look like a property.

PR #28, opened 2018-04-16
initial variable step size ("performance") representation for magenta.js

iansimon:


iansimon:
I'm also going to include a `Performance.fromNoteSequence` method but wanted to break things up into smaller pull requests where possible.  I could also imagine having this be a `DataConverter` but I think it makes more sense to have an explicit intermediate representation that can be used by multiple `DataConverter`s.

PR #32, opened 2018-04-18
create control signals from JSON spec (besides actual chords)

iansimon:
How's this?  Creating control signals from JSON spec is a little awkward because we need the actual chords for a chord progression.

### DIFF
--- a/core/src/controls.ts
+++ b/core/src/controls.ts
@@ -23,14 +23,14 @@ export interface BinaryCounterSpec {
   type: 'BinaryCounter';
   args: BinaryCounterArgs;
 }
-export type BinaryCounterUserArgs = {numSteps:number};
+export type BinaryCounterUserArgs = {};
 
 export type ChordProgressionArgs = {encoderType: ChordEncoderType};
 export interface ChordProgressionSpec {
   type: 'ChordProgression';
   args: ChordProgressionArgs;
 }
-export type ChordProgressionUserArgs = {chords:string[], stepsPerChord:number};
+export type ChordProgressionUserArgs = {chords:string[]};
 
 export type ControlSignalSpec = BinaryCounterSpec | ChordProgressionSpec;
 export type ControlSignalUserArgs = BinaryCounterUserArgs |
@@ -61,7 +61,7 @@ export function controlSignalFromSpec(spec: ControlSignalSpec) {
  */
 export abstract class ControlSignal<T> {
   readonly depth: number;     // Size of final output dimension.
-  abstract getTensors(args: T): tf.Tensor2D;
+  abstract getTensors(numSteps: number, args: T): tf.Tensor2D;
 
   constructor(depth: number) {
     this.depth = depth;
@@ -83,14 +83,14 @@ export class BinaryCounter extends ControlSignal<BinaryCounterUserArgs> {
    * @param numSteps Number of steps of binary counter values to return.
    * @returns A 2D tensor of binary counter values.
    */
-  getTensors(args: BinaryCounterUserArgs) {
-    const buffer = tf.buffer([args.numSteps, this.depth]);
-    for (let step = 0; step < args.numSteps; ++step) {
+  getTensors(numSteps: number, args: BinaryCounterUserArgs) {
+    const buffer = tf.buffer([numSteps, this.depth]);
+    for (let step = 0; step < numSteps; ++step) {
       for (let i = 0; i < this.depth; ++i) {
         buffer.set(Math.floor(step / Math.pow(2, i)) % 2 ? 1.0 : -1.0, step, i);
       }
     }
-    return buffer.toTensor().as2D(args.numSteps, this.depth);
+    return buffer.toTensor().as2D(numSteps, this.depth);
   }
 }
 
@@ -110,17 +110,17 @@ export class ChordProgression extends ControlSignal<ChordProgressionUserArgs> {
 
   /**
    * Get the tensor of control values. The returned tensor will be the encoded
-   * chord progression with each chord lasting the specified number of steps.
+   * chord progression over the requested number of steps.
    *
+   * @param numSteps Number of steps to use.
    * @param chords An array of chord symbol strings.
-   * @param stepsPerChord The number of steps each chord lasts.
    * @returns A 2D tensor containing the encoded chord progression.
    */
-  getTensors(args: ChordProgressionUserArgs) {
-    const totalSteps = args.chords.length * args.stepsPerChord;
+  getTensors(numSteps: number, args: ChordProgressionUserArgs) {
     const encodedChords = args.chords.map(chord => this.encoder.encode(chord));
-    const repeatedChords = encodedChords.map(ec =>
-        tf.tile(tf.expandDims(ec, 0), [args.stepsPerChord, 1]));
-    return tf.concat(repeatedChords, 0).as2D(totalSteps, this.depth);
+    const indices = Array.from(Array(numSteps).keys()).map(
+        step => Math.floor(step * encodedChords.length / numSteps));
+    return tf.stack(
+        indices.map(i => encodedChords[i])).as2D(numSteps, this.depth);
   }
 }
--- a/core/src/controls.ts
+++ b/core/src/controls.ts
@@ -97,7 +97,6 @@ export class BinaryCounter extends ControlSignal<BinaryCounterUserArgs> {
 /**
  * Chord progression control signal.
  *
- * @param numSteps The length of the control signal.
  * @param encoderType The chord encoder type to use.
  */
 export class ChordProgression extends ControlSignal<ChordProgressionUserArgs> {
--- a/core/src/controls.ts
+++ b/core/src/controls.ts
@@ -25,7 +25,7 @@ export interface BinaryCounterSpec {
   type: 'BinaryCounter';
   args: BinaryCounterArgs;
 }
-export type BinaryCounterUserArgs = {};
+export type BinaryCounterUserArgs = void;
 
 export type ChordProgressionArgs = {
   encoderType: ChordEncoderType
@@ -67,9 +67,9 @@ export function controlSignalFromSpec(spec: ControlSignalSpec) {
  *
  * @param depth The size of the control tensor at each step.
  */
-export abstract class ControlSignal<T> {
+export abstract class ControlSignal<A extends ControlSignalUserArgs> {
   readonly depth: number;  // Size of final output dimension.
-  abstract getTensors(numSteps: number, args: T): tf.Tensor2D;
+  abstract getTensors(numSteps: number, args: A): tf.Tensor2D;
 
   constructor(depth: number) { this.depth = depth; }
 }
--- a/core/src/controls.ts
+++ b/core/src/controls.ts
@@ -18,23 +18,29 @@
 import * as tf from '@tensorflow/tfjs';
 import {ChordEncoder, ChordEncoderType, chordEncoderFromType} from './chords';
 
-export type BinaryCounterArgs = {numBits: number};
+export type BinaryCounterArgs = {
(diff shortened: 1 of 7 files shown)


core/src/controls.ts
```diff
@@ -74,8 +106,9 @@ export class BinaryCounter extends ControlSignal {
 export class ChordProgression extends ControlSignal {
   readonly encodedChords: tf.Tensor1D[];
 
-  constructor(numSteps: number, chords: string[], encoder: ChordEncoder) {
-    super(numSteps, encoder.depth);
+  constructor(args: ChordProgressionArgs, chords: string[]) {
```
adarob: I was thinking that the constructor would only take the args so each model would have a single ControlSignal object for its lifetime. Then on each call, you would pass in an arg (e.g., a chord string) to generate a list of tensors for all steps, rather than calling a method for each step. This would be in line with how the DataConverter works.

core/src/controls.ts
```diff
@@ -16,7 +16,39 @@
  */
 
 import * as tf from '@tensorflow/tfjs';
-import {ChordEncoder} from './chords';
+import {ChordEncoderType, chordEncoderFromType} from './chords';
+
+export type BinaryCounterArgs = {numSteps: number, numBits: number};
```
adarob: I like how you have the spec set up, but it differs from how we're defining it in the DataConverter. Can you make the two consistent however you prefer?

core/src/controls.ts
iansimon: But what type should that arg be?  It works out in DataConverter as it always operates on a NoteSequence, but a) it's weird to have an extra function call for BinaryCounter that provides no information, and b) the fact is, different control signals cannot be treated equivalently as different DataConverters can (from the model's point of view); some control signals need user input, and some don't.

adarob: My thinking was that you would have a method like  `getTensors(args:ControlSequenceArgs)`. 
Then in MusicRNN you would have `continueSequence(sequence: magenta.INoteSequence, steps: number, temperature?: number, controlArgs?: ControlArgs)`.

How do you pass in the conditioning in your current setup?

core/src/controls.ts
iansimon: Done.

core/src/data.ts
```diff
@@ -44,50 +44,65 @@ const DEFAULT_DRUM_PITCH_CLASSES: number[][] = [
...
+export interface TrioConverterSpec {
+  type: 'TrioConverter';
+  args: TrioConverterArgs;
+}
+export interface DrumsOneHotConverterSpec {
+  type: 'DrumsOneHotConverter';
+  args: DrumsConverterArgs;
+}
+
 /**
```
adarob: Doc is no longer true, but I think it should be. Can we keep an interface that all of the specs extend so that they have to have a string type and a BaseConverterArgs?

core/src/data.ts
```diff
@@ -44,50 +44,65 @@ const DEFAULT_DRUM_PITCH_CLASSES: number[][] = [
...
+    case 'DrumsOneHotConverter': return new DrumsOneHotConverter(spec.args);
+    default: throw new Error(`Unknown DataConverter type: ${spec}`);
   }
 }
 
 /**
  * Abstract DataConverter class for converting between `Tensor` and
  * `NoteSequence` objects.
  */
+export interface BaseConverterArgs {
```
adarob: Not sure if this is the best place for this, especially with the current docstring above it.

core/src/data.ts
iansimon: If I set it up that way, the converterFromSpec function doesn't type check and I have to cast all of the different arg types.  I'm pretty sure the way I did it is idiomatic TypeScript.

iansimon: It's what they describe here if you search for "discriminated union": http://www.typescriptlang.org/docs/handbook/advanced-types.html

core/src/data.ts
iansimon: I'll move it just above the docstring.

core/src/controls.ts
```diff
@@ -16,82 +16,118 @@
...
+        buffer.set(Math.floor(step / Math.pow(2, i)) % 2 ? 1.0 : -1.0, step, i);
+      }
     }
-
-    const buffer = tf.buffer([this.depth]);
-    for (let i = 0; i < this.depth; ++i) {
-      buffer.set(Math.floor(step / Math.pow(2, i)) % 2 ? 1.0 : -1.0, i);
-    }
-    return buffer.toTensor().as1D();
+    return buffer.toTensor().as2D(numSteps, this.depth);
```
adarob: Should it be an array of 1D? Is this just going to have to be split later?

core/src/data.ts
adarob: Okay. Let's move this docstring and the expert type above the specs.

core/src/data.ts
```diff
@@ -44,46 +44,65 @@ const DEFAULT_DRUM_PITCH_CLASSES: number[][] = [
...
+    case 'TrioConverter':
+      return new TrioConverter(spec.args);
+    case 'DrumsOneHotConverter':
+      return new DrumsOneHotConverter(spec.args);
+    default:
+      throw new Error(`Unknown DataConverter type: ${spec}`);
   }
 }
 
+export interface BaseConverterArgs {
```
adarob: Add a docstring.

core/src/data.ts
adarob: Just to verify, there is no way to make sure these args are always BaseConverterArgs?

music_rnn/demo/index.ts
```diff
@@ -155,15 +155,16 @@ async function runMelodyRnn() {
...
+          {type: 'BinaryCounter', args: {numBits: 6}}));
   await drumsRnn.initialize();
 
   const qns = magenta.Sequences.quantizeNoteSequence(DRUMS_NS, 1);
   writeNoteSeqs('drums-cont-inputs', [qns]);
   const start = performance.now();
-  const continuation = await drumsRnn.continueSequence(
-      qns, 20, null,
-      new magenta.controls.BinaryCounter(qns.notes.length + 20, 6));
+  const continuation = await drumsRnn.continueSequence(qns, 20, null, {});
```
adarob: Do we need to pass in {}?

music_rnn/demo/index.ts
```diff
@@ -155,15 +155,16 @@ async function runMelodyRnn() {
 }
 
 async function runDrumsRnn() {
-  const drumsRnn = new MusicRNN(DRUMS_CHECKPOINT, null, 32);
+  const drumsRnn = new MusicRNN(
+      DRUMS_CHECKPOINT, null, 32,
+      magenta.controls.controlSignalFromSpec(
```
adarob: Can we set this up to automatically load from the checkpoint or is that a later PR?

music_rnn/src/model.ts
```diff
@@ -27,10 +27,11 @@ const CELL_FORMAT = 'multi_rnn_cell/cell_%d/basic_lstm_cell/';
  *
  * A MusicRNN is an LSTM-based language model for musical notes.
  */
-export class MusicRNN {
+export class MusicRNN<T extends magenta.controls.ControlSignalUserArgs> {
```
adarob: Does T have to be defined when the MusicRNN is created? If so, does that mean we have to change how we create MusicRNNs from a checkpoint where the T is defined in the checkpoint?

music_rnn/src/model.ts
```diff
@@ -137,8 +140,7 @@ export class MusicRNN {
    */
```
adarob: Add controlSignalArgs

music_rnn/src/model.ts
adarob: Maybe we should use a different name than "T" to make it more clear that this relates to the control sequence type.

music_rnn/src/model.ts
```diff
@@ -191,8 +198,9 @@ export class MusicRNN {
         samples.push(nextInput.as1D().toBool());
       }
 
-      if (controlSignal) {
-        const control = controlSignal.getTensor(i + 1);
+      if (controls) {
+        const control =
+            controls.slice([i, 0], [1, controlSize]).as2D(1, controlSize);
```
adarob: It should be more efficient if we just return a list for the controls instead of a 2D tensor.

core/src/data.ts
iansimon: The super(args) call wouldn't typecheck if they weren't.

core/src/controls.ts
```diff
@@ -16,82 +16,118 @@
...
    */
-  getTensor(step: number) {
-    if (step >= this.numSteps) {
-      throw new RangeError(`Step out of range: ${step} >= ${this.numSteps}`);
+  getTensors(numSteps: number, args: BinaryCounterUserArgs) {
+    const buffer = tf.buffer([numSteps, this.depth]);
+    for (let step = 0; step < numSteps; ++step) {
+      for (let i = 0; i < this.depth; ++i) {
+        buffer.set(Math.floor(step / Math.pow(2, i)) % 2 ? 1.0 : -1.0, step, i);
+      }
```
teropa: I think this may have an off-by-one issue, as [the binary counter is tracking the meter of the next step](https://github.com/tensorflow/magenta/blob/master/magenta/music/encoder_decoder.py#L522-L526).

iansimon: I guess it depends how this is supposed to align with the input sequence.  With chord conditioning, at each step the model should be fed melody[t-1] and chord[t], and output melody[t].  I thought it's the same for the binary counter...

teropa: Yeah that sounds about right, similar to chords the binary counter is one step ahead of the input. If that's how this is set up then should be fine :)

music_rnn/src/model.ts
iansimon: T is inferred based on the ControlSignal passed to the MusicRNN constructor.  I think as long as the JSON is a valid ControlSignalSpec everything should check out.  There's a similar step that happens in the initialize() method of MusicRNN, where the JSON response is assumed to be a ConverterSpec.

iansimon: I couldn't find anything about this in a TypeScript style guide; in most languages T is pretty idiomatic here (or sometimes K for keys, V for values, etc.).  Maybe I'll use A for args.

music_rnn/src/model.ts
iansimon: Talked offline, leaving 2D tensor for now but may change in multiple places later.

core/src/controls.ts
iansimon: Talked offline, will revisit later.

music_rnn/src/model.ts
iansimon: Done.

music_rnn/demo/index.ts
```diff
@@ -164,7 +164,7 @@ async function runDrumsRnn() {
   const qns = magenta.Sequences.quantizeNoteSequence(DRUMS_NS, 1);
   writeNoteSeqs('drums-cont-inputs', [qns]);
   const start = performance.now();
-  const continuation = await drumsRnn.continueSequence(qns, 20, null, {});
+  const continuation = await drumsRnn.continueSequence(qns, 20, null);
```
adarob: Leave off the null?

music_rnn/src/model.ts
```diff
@@ -137,10 +137,11 @@ export class MusicRNN<T extends magenta.controls.ControlSignalUserArgs> {
    * @param steps How many steps to continue.
    * @param temperature The softmax temperature to use when sampling from the
```
adarob: While you're at it, please add the (Optional) prefix to temperature.

iansimon: Done.

PR #33, opened 2018-04-19
Switch to tfjs-core and fix compatibility issues with newer version of tfjs.

adarob:
This will decrease the size of our bundle since we don't need tfjs-layers.


core/src/controls.ts
```diff
@@ -69,7 +69,7 @@ export function controlSignalFromSpec(spec: ControlSignalSpec) {
  */
 export abstract class ControlSignal<A extends ControlSignalUserArgs> {
   readonly depth: number;  // Size of final output dimension.
-  abstract getTensors(numSteps: number, args: A): tf.Tensor2D;
+  abstract getTensors(numSteps: number, args?: A): tf.Tensor2D;
```
iansimon: This isn't right, because then getTensors in the subclasses has to be able to handle the case of missing args.

music_rnn/src/model.ts
```diff
@@ -204,10 +206,9 @@ export class MusicRNN<A extends magenta.controls.ControlSignalUserArgs> {
         break;
       }
 
-      if (controls) {
-        const control =
-            controls.slice([i + 1, 0], [1, controlSize]).as2D(1, controlSize);
-        nextInput = nextInput.concat(control.as2D(1, -1), 1);
+      if (splitControls) {
+        const control = splitControls[i] as tf.Tensor2D;
```
iansimon: Should be i + 1.

core/src/controls.ts
adarob: Done.

music_rnn/src/model.ts
adarob: Done.

PR #36, opened 2018-04-20
support control signals in MusicVAE

iansimon:


music_rnn/src/model.ts
```diff
@@ -52,6 +52,8 @@ export class MusicRNN<A extends magenta.controls.ControlSignalUserArgs> {
    * `NoteSequence` and `Tensor` objects. If not provided, a `converter.json`
```
adarob: Please add (Optional) here too for consistency.

adarob: Can you add an example to the demo either here or in a follw-up PR?

iansimon: I'll do another PR where I add examples with chord-conditioning in both music_rnn and music_vae.

iansimon: Done.

Issue #11, opened 2018-04-05
Switch to new TensorFlow.js checkpoints.


adarob:

adarob:
Include checkpoints for:
- [ ] Melody BasicRNN
- [ ] Melody AttentionRNN (w/o special inputs)
- [ ] DrumsRNN
- [ ] ImprovRNN
- [ ] PerformanceRNN
- [ ] MusicVAE: 2-bar Melody
- [ ] MusicVAE: 4-bar Melody
- [ ] MusicVAE: 2-bar Drums (hi kl)
- [ ] MusicVAE: 2-bar Drums (lo kl)
- [ ] MusicVAE: 4-bar Drums
- [ ] MusicVAE: 4-bar Trio
- [ ] MusicVAE: 16-bar Trio

Quantize the larger ones. We should also keep the original TF checkpoints in the folders for future reference.

adarob:
This is partially blocked by adding checkpoint quantization support for tfjs, which is now under review:
https://github.com/tensorflow/tfjs-core/pull/965
https://github.com/tensorflow/tfjs-converter/pull/87

adarob:
Closed by #55

PR #37, opened 2018-04-20
Add support for new tfjs checkpoints to MusicRNN along with a checkpiont dump script.

adarob:
Our own script was needed since tfjs-converter is not supporting checkpoint-only conversions (fyi, @nsthorat 😎).


adarob:
Partially solves #11

PR #39, opened 2018-04-20
Switch MusicVAE to new checkpoints.

adarob:


music_vae/src/model.ts
```diff
@@ -704,7 +705,7 @@ class MusicVAE<A extends magenta.controls.ControlSignalUserArgs> {
     const interpZs = tf.tidy(() => this.getInterpolatedZs(inputZs, numInterps));
     inputZs.dispose();
 
-    const outputSequenes = this.decode(interpZs, controlSignalArgs);
+    const outputSequenes = this.decode(interpZs, undefined, controlSignalArgs);
```
iansimon: Why not have the method take a temperature argument?

music_vae/src/model.ts
```diff
@@ -724,7 +725,7 @@ class MusicVAE<A extends magenta.controls.ControlSignalUserArgs> {
       let inputTensors = tf.stack(inputSequences.map(
           t => this.dataConverter.toTensor(t) as tf.Tensor2D)) as tf.Tensor3D;
 
-      if (this.controlSignal) {
+      if (!isNullOrUndefined(this.controlSignal)) {
```
iansimon: What's different about this check?

adarob: Nothing in this case. I tried this when I wasn't sure of the original problem. Reverted.

PR #41, opened 2018-04-22
add minimal config for DataConverter/ControlSignal and fix some type issues

iansimon:


core/src/config.ts
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+import * as controls from './controls';
+import * as data from './data';
+
+// TODO(iansimon): support multiple control signals
+
+export interface ConfigSpec {
+  dataConverter: data.ConverterSpec;
```
adarob: I think we also need a section for model args (like attention length).

I'm wondering if we should actually call this ModelSpec and have it include the model type (for now, MusicRNN | MusicVAE) as well as model args. Then we can just have a modelFromCheckpoint function. We would move the loading of checkpoints out of the model classes and make the vars be an argument.

I tmay also be good to move all of the data and control specs into here, but that's not as important.

iansimon: Are you imagining the modelFromCheckpoint function returns an object of type MusicRNN | MusicVAE?  I'm not sure how I feel about that; I think there should be one function (possibly just new()) to instantiate one and one to instantiate the other, since the interfaces for using the models are so different.

What else is there (besides attention length) that isn't part of a DataConverter or ControlSignal?  Maybe attention should actually be a ControlSignal...

PR #45, opened 2018-04-24
Update checkpoint paths and add demo index page.

adarob:


music/demos/build-demos.sh
```diff
@@ -26,7 +26,7 @@
 if [[ -n $1 ]]; then
   demos=( "${1}" )
 else
-  demos=( *.html )
+  demos=( *.ts )
```
cghawthorne: My thought for using .html was that we might eventually have demos with multiple .ts files, but they'll probably be launched from one html file. But I see the problem with index.html now. We could either specifically exclude it or just go with this approach until we need something more complicated. :-)

music/demos/music_rnn.ts
```diff
@@ -151,7 +152,8 @@ async function runMelodyRnn() {
   console.log(tf.memory());
 }
 
-async function runDrumsRnn() {
+async function
+runDrumsRnn() {
```
cghawthorne: why is this on 2 lines?

music/demos/music_vae.ts
```diff
@@ -293,8 +294,8 @@ runTrio() {
 }
 
 // TODO(adarob): Switch to magenta/core function once implemented.
-function concatNoteSequences(seqs: mm.INoteSequence[],
-    individualDuration: number) {
+function
+concatNoteSequences(seqs: mm.INoteSequence[], individualDuration: number) {
```
cghawthorne: ditto

adarob: Done.

music/demos/music_rnn.ts
adarob: Done.

PR #47, opened 2018-04-24
Switch back to tfjs-core and take advantage of updates.

adarob:


music/src/core/data.ts
```diff
@@ -69,8 +71,8 @@ export interface DrumsOneHotConverterSpec {
  * @property args Map containing values for argments to the constructor of the
  * `DataConverter` class specified above.
  */
-export type ConverterSpec = MelodyConverterSpec | DrumsConverterSpec |
-    DrumRollConverterSpec | TrioConverterSpec | DrumsOneHotConverterSpec;
+export type ConverterSpec = MelodyConverterSpec|DrumsConverterSpec|
```
iansimon: Did something happen with our clang-format configuration?  It used to put spaces here...

adarob: Discussed offline.

PR #52, opened 2018-04-24
refactor control signals into chords and auxiliary inputs

iansimon:


music/src/core/aux_inputs.ts
```diff
@@ -0,0 +1,93 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as tf from '@tensorflow/tfjs-core';
+
+/**
+ * This file contains "auxiliary" model inputs, which can be appended to the
```
adarob: I think these should probably go above the imports.

music/src/music_rnn/model.ts
```diff
@@ -155,11 +162,12 @@ export class MusicRNN<A extends controls.ControlSignalUserArgs> {
...
    * @param temperature (Optional) The softmax temperature to use when sampling
    * from the logits. Argmax is used if not provided.
-   * @param controlSignalArgs (Optional) Arguments for creating control tensors.
+   * @param chordProgression (Optional) Chord progression to use as
+   * conditioning.
    */
   async continueSequence(
       sequence: INoteSequence, steps: number, temperature?: number,
-      controlSignalArgs?: A): Promise<INoteSequence> {
+      chordProgression = [constants.NO_CHORD]): Promise<INoteSequence> {
```
adarob: Is NO_CHORD a valid input to a chord encoder?
I was actually thinking this would be undefined by default, and then we would throw an error if a chordEncoder existed but the user didn't pass in a chordProgression, and vice-versa. This is going to be confusing for people, so we need to make the failures really clear.

music/src/music_rnn/model.ts
```diff
@@ -226,7 +244,10 @@ export class MusicRNN<A extends controls.ControlSignalUserArgs> {
       }
 
       if (splitControls) {
-        nextInput = nextInput.concat(splitControls[i + 1], 1);
+        nextInput = tf.concat([splitControls[i + 1], nextInput], 1);
```
adarob: We should do a single concat for performance reasons.
You should build up the array here and then call concat once.

music/src/music_rnn/model.ts
```diff
@@ -57,20 +60,24 @@ export class MusicRNN<A extends controls.ControlSignalUserArgs> {
    * converting between `NoteSequence` and `Tensor` objects. If not provided, a
    * `converter.json` file must exist within the checkpoint directory specifying
    * the type and args for the correct `DataConverter`.
-   * @param controlSignal (Optional) A `ControlSignal` object that produces
-   * control tensors that will be appended to model inputs.
+   * @param chordEncoder (Optional) A `ChordEncoder` object that converts chord
+   * symbol strings to model input tensors.
+   * @param auxInputs (Optional) An array of `AuxiliaryInput` objects that
```
adarob: ditto

music/src/music_rnn/model.ts
```diff
@@ -57,20 +60,24 @@ export class MusicRNN<A extends controls.ControlSignalUserArgs> {
    * converting between `NoteSequence` and `Tensor` objects. If not provided, a
    * `converter.json` file must exist within the checkpoint directory specifying
    * the type and args for the correct `DataConverter`.
-   * @param controlSignal (Optional) A `ControlSignal` object that produces
-   * control tensors that will be appended to model inputs.
+   * @param chordEncoder (Optional) A `ChordEncoder` object that converts chord
+   * symbol strings to model input tensors.
```
adarob: Mention that these are concatenated before(after?) the input tensor from the data converter.

music/src/music_vae/model.ts
```diff
@@ -18,7 +18,8 @@
 import * as tf from '@tensorflow/tfjs-core';
```
adarob: Same reviewer comments as for MusicRNN apply here as well.

iansimon: BTW I didn't add auxiliary inputs for MusicVAE because none of our models actually use them.

adarob: To clarify, I meant that to say that lots of the requests I made above are relevant here as well, but I didn't want to point them all out :)

music/src/music_rnn/model.ts
iansimon: NO_CHORD is a valid input to a chord encoder.  It's a question of what should happen if chords are not passed to a chord-conditioned model.  I think you're right, it should error.

music/src/music_vae/model.ts
iansimon: Yes :)  I think because no auxiliary inputs it's just changing the chord progression default, which I fixed.

music/src/core/aux_inputs.ts
iansimon: Done.

music/src/music_rnn/model.ts
iansimon: Done.

music/src/music_rnn/model.ts
iansimon: Done.

music/src/music_rnn/model.ts
iansimon: Done.

music/src/music_rnn/model.ts
```diff
@@ -155,13 +163,18 @@ export class MusicRNN<A extends controls.ControlSignalUserArgs> {
...
    */
   async continueSequence(
       sequence: INoteSequence, steps: number, temperature?: number,
-      controlSignalArgs?: A): Promise<INoteSequence> {
+      chordProgression?: string[]): Promise<INoteSequence> {
     sequences.assertIsQuantizedSequence(sequence);
 
+    if (this.chordEncoder && !chordProgression) {
+      throw new Error('Chord progression expected but not provided.');
+    }
```
adarob: Also add an error if one is provided but not expected.

music/src/music_rnn/model.ts
```diff
@@ -170,10 +183,18 @@ export class MusicRNN<A extends controls.ControlSignalUserArgs> {
...
       const length: number = inputs.shape[0];
       const outputSize: number = inputs.shape[1];
-      const controls = this.controlSignal ?
-          this.controlSignal.getTensors(length + steps, controlSignalArgs) :
+      const controls = this.chordEncoder ?
+          this.chordEncoder.encodeProgression(
+              chordProgression, length + steps) :
           undefined;
-      const samples = this.sampleRnn(inputs, steps, temperature, controls);
+      const auxInputs = this.auxInputs ?
```
adarob: The concat and then split is wasteful.

iansimon: They're on different axes.

music/src/music_rnn/model.ts
iansimon: Done.

PR #63, opened 2018-04-27
make model constructors synchronous

iansimon:


music/src/music_rnn/model.ts
```diff
@@ -142,13 +95,43 @@ export class MusicRNN {
...
   /**
    * Loads variables from the checkpoint and instantiates the `Encoder` and
    * `Decoder`.
    */
   async initialize() {
     this.dispose();
 
+    if (!this.spec) {
+      await fetch(`${this.checkpointURL}/config.json`)
+          .then((response) => response.json())
```
adarob: We should throw an error if the file isn't found.

iansimon: Will fix in subsequent PR.

PR #67, opened 2018-04-30
Fixes for playback.

adarob:
* Player handles qpm correctly.
* Player respects totalTime for resolution of promise.
* Adds unquantizeSequence function.

### DIFF
--- a/music/package.json
+++ b/music/package.json
@@ -9,6 +9,7 @@
   "dependencies": {
     "@tensorflow/tfjs-core": "^0.8.5",
     "midiconvert": "^0.4.4",
+    "node": "^10.0.0",
     "protobufjs": "^6.8.6",
     "tonal": "^1.1.3",
     "tone": "^0.12.80",
--- a/music/src/core/player.ts
+++ b/music/src/core/player.ts
@@ -15,12 +15,19 @@
  * =============================================================================
  */
 import * as Tone from 'tone';
+import {isNullOrUndefined} from 'util';
 
 import {INoteSequence, NoteSequence} from '../protobuf/index';
 
 import {sequences} from '.';
-import * as constants from './constants';
 import {DEFAULT_DRUM_PITCH_CLASSES} from './data';
+
+const DRUM_PITCH_TO_CLASSS = new Map<number, number>();
+for (let c = 0; c < DEFAULT_DRUM_PITCH_CLASSES.length; ++c) {  // class
+  DEFAULT_DRUM_PITCH_CLASSES[c].forEach((p) => {
+    DRUM_PITCH_TO_CLASSS.set(p, c);
+  });
+}
 
 const kick = new Tone.MembraneSynth().toMaster();
 const tomLow = new Tone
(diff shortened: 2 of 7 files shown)


music/package.json
```diff
@@ -9,6 +9,7 @@
   "dependencies": {
     "@tensorflow/tfjs-core": "^0.8.5",
     "midiconvert": "^0.4.4",
+    "node": "^10.0.0",
```
iansimon: Why do we need node?

music/src/core/player.ts
```diff
@@ -15,13 +15,20 @@
...
 import * as Tone from 'tone';
+import {isNullOrUndefined} from 'util';
 
 import {INoteSequence, NoteSequence} from '../protobuf/index';
 
 import {sequences} from '.';
-import * as constants from './constants';
 import {DEFAULT_DRUM_PITCH_CLASSES} from './data';
 
+const DRUM_PITCH_TO_CLASSS = new Map<number, number>();
```
iansimon: DRUM_PITCH_TO_CLASS

music/package.json
cghawthorne: what is this needed for? if it's just for type definitions, it could be a dev dependency.

music/src/core/sequences.ts
```diff
@@ -325,3 +334,58 @@ export function assertIsAbsoluteQuantizedSequence(ns: INoteSequence) {
...
+  } else {
+    qpm = (qns.tempos && qns.tempos.length > 0) ?
+        ns.tempos[0].qpm :
+        constants.DEFAULT_QUARTERS_PER_MINUTE;
+  }
+
+  const stepToSeconds = (step: number) =>
+      step / ns.quantizationInfo.stepsPerQuarter * (60 / qpm);
+  ns.totalTime = stepToSeconds(ns.totalQuantizedSteps);
+  ns.totalQuantizedSteps = undefined;
```
cghawthorne: I don't think we need to clear out quantized information when adding unquantized information. Having both is only problematic if we're trying to do something like concatenate quantized sequences.

music/src/core/sequences.ts
```diff
@@ -325,3 +334,58 @@ export function assertIsAbsoluteQuantizedSequence(ns: INoteSequence) {
...
+    n.endTime = stepToSeconds(n.quantizedEndStep);
+    n.quantizedStartStep = undefined;
+    n.quantizedEndStep = undefined;
+
+    // Extend sequence if necessary.
+    ns.totalTime = Math.max(ns.totalTime, n.endTime);
+  });
+
+  // Also quantize control changes and text annotations.
+  ns.controlChanges.concat(ns.textAnnotations).forEach(event => {
```
cghawthorne: for completeness: time signatures, key signatures, pitch bends, section annotations. tempos too, but those are taken care of above.

it might be useful to have a function that returns a concatenated list of these timed events for any given sequence.

music/package.json
adarob: Done.

adarob: Done.

music/src/core/player.ts
adarob: DONE

music/src/core/sequences.ts
adarob: Done.

music/src/core/sequences.ts
adarob: done for those event types that have quantized steps

PR #70, opened 2018-05-01
Refactor drums synth for simpler documentation.

adarob:


music/src/core/player.ts
```diff
@@ -22,89 +22,116 @@ import {INoteSequence, NoteSequence} from '../protobuf/index';
...
+
+  static getInstance() {
+    if (!DrumKit.instance) {
+      DrumKit.instance = new DrumKit();
+    }
+    return DrumKit.instance;
+  }
 
+  public playNote(pitch: number, time: number) {
+    [() => this.kick.triggerAttackRelease('C2', '8n', time),
```
cghawthorne: Could this array be defined somewhere else (maybe as a member variable in the constructor)? I think as written, it will have to redefine the array for every call.

adarob: Done

PR #73, opened 2018-05-03
add multitrack data converter

iansimon:


music/src/core/data.ts
```diff
@@ -493,3 +501,242 @@ export class TrioConverter extends DataConverter {
...
+    // note-on, note-off, time-shift, velocity-change, program-select, end-token
+
+    this.numPitches = constants.MAX_MIDI_PITCH - constants.MIN_MIDI_PITCH + 1;
+    this.performanceEventDepth =
+        2 * this.numPitches + this.totalSteps + this.numVelocityBins;
+
+    // Include an extra "program" for drums.
+    this.numPrograms =
+        constants.MAX_MIDI_PROGRAM - constants.MIN_MIDI_PROGRAM + 2;
+    this.depth = this.performanceEventDepth + this.numPrograms + 1;
```
adarob: maybe add a comment that the 1 is for end token

music/src/core/data.ts
```diff
@@ -493,3 +501,242 @@ export class TrioConverter extends DataConverter {
...
+
+    const instruments =
+        new Set(noteSequence.notes.map(note => note.instrument));
+    const tracks =
+        Array.from(instruments)
+            .map(
+                instrument => performance.Performance.fromNoteSequence(
+                    noteSequence, this.totalSteps, this.numVelocityBins,
+                    instrument));
+    const sortedTracks = tracks.sort(
```
adarob: A comment for the sort order would be helpful here.

music/src/core/data.ts
```diff
@@ -493,3 +501,242 @@ export class TrioConverter extends DataConverter {
...
+
+    // Convert tracks to tensors then concatenate.
+    return tf.tidy(
+        () => tf.concat(
+            sortedTracks.map((track) => this.trackToTensor(track)), 0));
+  }
+
+  private tokensToTrack(tokens: Int32Array) {
+    // Trim to end token.
+    const idx = tokens.indexOf(this.depth - 1);
```
adarob: I think it would be useful to store the end token as a property for readability.

music/src/core/data.ts
iansimon: Used explicit endToken property.

music/src/core/data.ts
iansimon: Done.

music/src/core/data.ts
iansimon: Done.

PR #76, opened 2018-05-05
use segment lengths in hierarchical encoder

iansimon:


music/src/music_vae/model.ts
```diff
@@ -99,10 +100,11 @@ class BidirectonalLstmEncoder extends Encoder {
   /**
    * Encodes a batch of sequences.
    * @param sequence The batch of sequences to be encoded.
+   * @param segmentLengths Unused for this encoder.
    * @returns A batch of concatenated final LSTM states, or the `mu` if `muVars`
    * is known.
    */
-  encode(sequence: tf.Tensor3D) {
+  encode(sequence: tf.Tensor3D, segmentLengths?: number[]) {
```
adarob: assert segment lengths is undefined

iansimon: Done.

PR #79, opened 2018-05-10
support chord-conditioning for multitrack model

iansimon:


music/src/music_vae/model.ts
```diff
@@ -865,13 +867,30 @@ class MusicVAE {
     }
 
     if (this.chordEncoder) {
-      // TODO(iansimon): handle chords for variable-length segments
```
adarob: I think it would be more readable to put this and the variable-length code above into separate private methods.

adarob: Ideally it would also be used by the decoder.

music/src/music_vae/model.ts
```diff
@@ -908,17 +927,44 @@ class MusicVAE {
...
+      let controls: tf.Tensor2D = undefined;
+      if (this.chordEncoder) {
+        // Construct the chord-conditioning tensors. For models where each
+        // segment is a separate track, the same chords should be used for
+        // each segment. If an initial chord is specified by the data
+        // converter, use that chord for the first step only.
+        const numChordSteps = this.dataConverter.SEGMENTED_BY_TRACK ?
+            numSteps / numSegments :
+            numSteps;
+        const encodedChordProgression = this.dataConverter.INITIAL_CHORD ?
```
adarob: Is INITIAL_CHORD only used when SEGMENTED_BY_TRACK is used? If so, do we need both?

PR #86, opened 2018-05-22
initial support for soundfonts

iansimon:


music/src/core/player.ts
```diff
@@ -201,20 +227,35 @@ export class Player {
...
+  async loadSamples(seq: INoteSequence): Promise<void> {
+    await this.soundfont.loadSamples(seq.notes.map((note) => ({
+                                                     pitch: note.pitch,
+                                                     velocity: note.velocity,
+                                                     program: note.program,
+                                                     isDrum: note.isDrum
+                                                   })));
+  }
+
+  protected playNote(time: number, note: NoteSequence.INote) {
```
adarob: what happens if initialize hasn't been called?

music/demos/music_rnn.ts
```diff
@@ -144,9 +148,13 @@ function writeNoteSeqs(elementId: string, seqs: mm.INoteSequence[]) {
 }
 
 function createPlayer(seq: mm.INoteSequence) {
-  const player = new mm.Player();
+  const player = new mm.GMPlayer(SOUNDFONT_URL);
   const button = document.createElement('button');
   button.textContent = 'Play';
+  button.disabled = true;
+  player.initialize()
+      .then(() => player.loadSamples(seq))
```
adarob: shouldn't this be called inside "start" so the user doesn't have to remember to call it each time?

music/src/core/soundfont.ts
```diff
@@ -0,0 +1,377 @@
...
+export interface SampleInfo {
+  pitch: number;
+  velocity: number;
+}
+
+/**
+ * Specification for a sampled instrument. Samples must exist for all pitches
+ * between `minPitch` and `maxPitch` at all velocities in `velocities` (unless
+ * `velocities` is undefined, in which case only a single sample exists for each
+ * pitch. Each sample consists of a note sustained for `durationSeconds` then
```
adarob: missing close parens

music/src/core/soundfont.ts
```diff
@@ -0,0 +1,377 @@
...
+
+/**
+ * Sampled instrument. Must be initialized and samples must be pre-loaded using
+ * the `loadSamples` method before any notes can be played.
+ */
+export class Instrument {
+  private FADE_SECONDS = 0.1;
+
+  private readonly baseURL: string;
+  // tslint:disable-next-line:no-any
```
adarob: // tslint:disable-line on same line preferred here and elsewhere

music/src/core/soundfont.ts
```diff
@@ -0,0 +1,377 @@
...
+  private readonly buffers: any;
+
+  private initialized: boolean;
+
+  name: string;
```
adarob: If this list will always be the same as InstrumentSpec, maybe just do `spec: InstrumentSpec` as the instance variable

music/src/core/soundfont.ts
```diff
@@ -0,0 +1,377 @@
...
+    return bestVelocity;
+  }
+
+  /**
+   * Load samples necessary to play a set of pitch/velocity pairs. This must be
+   * called before any notes can be played.
+   *
+   * @param samples Array of pitch/velocity pairs.
+   */
+  async loadSamples(samples: SampleInfo[]) {
```
adarob: does this avoid reloading samples?

music/src/core/soundfont.ts
```diff
@@ -0,0 +1,377 @@
...
+   * Play a note using one of the samples.
+   *
+   * @param pitch Pitch of the note.
+   * @param velocity Velocity of the note.
+   * @param startTime Time at which to start playing the note.
+   * @param duration Length of the note in seconds.
+   */
+  playNote(
+      pitch: number, velocity: number, startTime: number, duration: number) {
+    if (!this.initialized) {
```
adarob: why not just initialize this lazily?

music/src/core/soundfont.ts
```diff
@@ -0,0 +1,377 @@
...
+    }
+    if (pitch < this.minPitch || pitch > this.maxPitch) {
+      console.log(`Pitch ${pitch} is outside the valid range for ${
+          this.name} (${this.minPitch}-${this.maxPitch})`);
+      return;
+    }
+
+    const name = this.sampleInfoToName(
+        {pitch, velocity: this.nearestVelocity(velocity)});
+    if (!this.buffers.has(name)) {
```
adarob: why not load samples lazily?

music/src/core/soundfont.ts
```diff
@@ -0,0 +1,377 @@
...
+   * @param duration Length of the note in seconds.
+   * @param program Program number to use for instrument lookup.
+   * @param isDrum Drum status to use for instrument lookup.
+   */
+  playNote(
+      pitch: number, velocity: number, startTime: number, duration: number,
+      program: number, isDrum: boolean) {
+    const instrument = isDrum ? 'drums' : program;
+    if (!this.initialized) {
+      throw new Error('SoundFont is not initialized.');
```
adarob: ditto on lazy loading

music/src/core/player.ts
```diff
@@ -201,20 +227,35 @@ export class Player {
...
-    if (this.currentPart) {
-      this.currentPart.stop();
-      this.currentPart = null;
-    }
-    Tone.Transport.clear(this.scheduledStop);
-    this.scheduledStop = null;
+/**
+ * A `NoteSequence` player based on Tone.js that uses SoundFont samples.
+ */
+export class GMPlayer extends BasePlayer {
```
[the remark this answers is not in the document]
iansimon: What about just renaming it SoundFontPlayer?  There's nothing about it that really requires General MIDI, it just uses program numbers to select different sampled instruments.

music/src/core/player.ts
```diff
@@ -201,20 +227,35 @@ export class Player {
...
-   */
-  stop() {
-    if (this.currentPart) {
-      this.currentPart.stop();
-      this.currentPart = null;
-    }
-    Tone.Transport.clear(this.scheduledStop);
-    this.scheduledStop = null;
+/**
+ * A `NoteSequence` player based on Tone.js that uses SoundFont samples.
```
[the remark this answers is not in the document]
iansimon: see below

music/src/core/player.ts
iansimon: In order to match the behavior of the other Player, the implication is that playNote will immediately schedule the note and does not depend on disk/network calls.  The Instrument will throw an error if the sample hasn't already been loaded.  I'll add a comment.

music/demos/music_rnn.ts
iansimon: Similar to above, my intention was to more closely match the behavior of the Player class, where "start" schedules everything immediately.

I could add another method loadSamplesAndStart...

music/src/core/soundfont.ts
iansimon: Like above, implication is that playNote schedules immediately.

music/src/core/soundfont.ts
iansimon: Same as above.

music/src/core/soundfont.ts
iansimon: Likewise :)

music/src/core/soundfont.ts
iansimon: Yes, it filters out samples for which we already have a buffer.  Will add a comment.

music/src/core/soundfont.ts
iansimon: Done.

music/src/core/soundfont.ts
iansimon: Done.

music/src/core/soundfont.ts
iansimon: I slightly prefer it this way to make accesses to those variables simpler.

music/demos/music_rnn.ts
adarob: I think this is pretty confusing since you have to call 2 methods every time you want to play something.  I'd argue you should either just lazily load the samples when you call start(), or load all of the samples in initialize().

music/src/core/player.ts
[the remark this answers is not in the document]
adarob: SGTM

music/src/core/player.ts
```diff
@@ -201,20 +224,37 @@ export class Player {
...
-  stop() {
-    if (this.currentPart) {
-      this.currentPart.stop();
-      this.currentPart = null;
-    }
-    Tone.Transport.clear(this.scheduledStop);
-    this.scheduledStop = null;
+/**
+ * A `NoteSequence` player based on Tone.js that uses SoundFont samples. The
+ * `loadSamples` method must be called before `start` so that the necessary
```
adarob: `loadSamples` is no longer *required* before `start`.  Can you also avoid having to explicitly call initialize (see below)?

I think you should expand this a bit more, maybe even including a bit of example usage. There are few things to note with regards to when there will be a delay.  For example:

> When you start playing a sequence, the SoundFont information and required samples must be downloaded, which will lead to a delay.
> To avoid this delay when starting playback, you can pre-initialize the class and load the required samples, e.g.:
> ```js
> player.initialize().then(() => player.loadSamples(seq));  
> ```
> Once this completes `player.start(seq)` will start without delay.

Also, if you lazily call initialize in `loadSamples`, `player.loadSamples(seq)` will be sufficient.

iansimon: Fixed comments.  Also, SoundFontPlayer no longer has an initialize method.

PR #87, opened 2018-05-24
refactor model demos and fix duplicate samples bug

iansimon:


music/demos/common.ts
```diff
@@ -0,0 +1,203 @@
...
+            })
+            .join(', ') +
+        ']';
+    seqWrap.appendChild(seqText);
+    seqWrap.appendChild(createPlayer(seq));
+    element.appendChild(seqWrap);
+  });
+}
+
+function createPlayer(seq: mm.INoteSequence) {
```
adarob: allow you to choose between player types? we should probably keep the non-multitrack ones on the original player

iansimon: Why?

iansimon: Switched back to original player for multitrack.  Was tempted to do trios too but that's the only sequence type that uses the synth bass...

iansimon: Oops, I meant all but multitrack.

music/src/core/sequences.ts
```diff
@@ -399,8 +399,8 @@ export function unquantizeSequence(qns: INoteSequence, qpm?: number) {
 /**
  * Assign instruments to the notes, pitch bends, and control changes of a
  * `NoteSequence` based on program numbers and drum status. All drums will be
- * assigned instrument 9 (and program 0). All non-drum events with the same
- * program number will be assigned to a single instrument.
+ * assigned the last instrument (and program 0). All non-drum events with the
```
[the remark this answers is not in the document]
iansimon: In MIDI files it's the "channel" (which isn't the same as instrument/program), but in NoteSequence we use isDrum for that.  In the to/from MIDI routines you can see how channel and isDrum are set.

PR #90, opened 2018-05-28
MusicVAE.interpolate() accepts 2 dimensions, for non-square outputs

hrldcpr:
This lets `MusicVAE.interpolate()` output not just squares, but arbitrary rectangles.

The `number | [number, number]` thing is maybe a little weird, but it seems like the best backwards-compatible way to accomplish this.


hrldcpr:
👍

PR #94, opened 2018-05-30
upgrade to latest midiconvert with MIDI writing bugfix

iansimon:


cghawthorne:
Can we update the dependencies in package.json?

PR #100, opened 2018-06-09
Add a Player with callback functionality.

psc-g:
This allows you to sync other browser events with the playing of notes, which facilitates building web interfaces.

I've added a demo to the demos folder. Do check it out if you can, as I think it's quite cool :).


music/src/core/player.ts
```diff
@@ -328,9 +328,89 @@ export class SoundFontPlayer extends BasePlayer {
 }
 
 /**
- * A `NoteSequence` player based on Tone.js that includes a click track.
+ * A base class for providing arbitrary callbacks for each note played.
  */
-export class PlayerWithClick extends Player {
+export class BasePlayerCallback {
```
adarob: make this an abstract class with abstract methods signatures

music/src/core/player.ts
```diff
@@ -328,9 +328,89 @@ export class SoundFontPlayer extends BasePlayer {
...
+
+  constructor(playClick?: boolean, callbackObject?: BasePlayerCallback) {
+    super();
+    if (!playClick) {
+      this.playClick = true;
+    } else {
+      this.playClick = playClick;
+    }
+    if (!callbackObject) {
+      this.callbackObject = new BasePlayerCallback();
```
adarob: just leave this as undefined if it's not provided and use conditionals below

music/src/core/player.ts
```diff
@@ -328,9 +328,89 @@ export class SoundFontPlayer extends BasePlayer {
...
+ * A `NoteSequence` player based on Tone.js that allows arbitrary callbacks
+ * when playing each note.
+ */
+export class PlayerWithCallback extends Player {
+  private playClick: boolean;
+  private callbackObject: BasePlayerCallback;
+
+  constructor(playClick?: boolean, callbackObject?: BasePlayerCallback) {
+    super();
+    if (!playClick) {
```
adarob: just use a default value in the signature

music/src/core/player.ts
```diff
@@ -328,9 +328,89 @@ export class SoundFontPlayer extends BasePlayer {
...
+
+/**
+ * A `NoteSequence` player based on Tone.js that allows arbitrary callbacks
+ * when playing each note.
+ */
+export class PlayerWithCallback extends Player {
+  private playClick: boolean;
+  private callbackObject: BasePlayerCallback;
+
+  constructor(playClick?: boolean, callbackObject?: BasePlayerCallback) {
```
adarob: add constructor docstring w/ args (eg, https://github.com/tensorflow/magenta-js/blob/455660cfa1de8e1d6300795e4cda209a0fc55038/music/src/music_vae/model.ts#L166)

music/src/index.ts
```diff
@@ -18,6 +18,8 @@
 import * as tf from '@tensorflow/tfjs-core';
 export {tf};
 
+export * from './core/constants';
```
adarob: I don't think you should be adding these here. Instead, you should use mm.constants.CONSTANT_NAME, etc.

music/src/core/player.ts
```diff
@@ -328,9 +328,89 @@ export class SoundFontPlayer extends BasePlayer {
...
+    } else {
+      this.playClick = playClick;
+    }
+    if (!callbackObject) {
+      this.callbackObject = new BasePlayerCallback();
+    } else {
+      this.callbackObject = callbackObject;
+    }
+  }
+  /**
```
adarob: add empty line

music/src/core/player.ts
```diff
@@ -328,9 +328,89 @@ export class SoundFontPlayer extends BasePlayer {
...
+    if (sequences.isQuantizedSequence(seq)) {
+      seq = sequences.unquantizeSequence(seq, qpm);
+    } else if (qpm) {
+      throw new Error('Cannot specify a `qpm` for a non-quantized sequence.');
+    }
+
+    this.currentPart = new Tone.Part(
+      (t: number, n: NoteSequence.INote) => {
+        if (this.playClick ||
+            (n.pitch != constants.LO_CLICK_PITCH &&
```
adarob: Why do you need to filter this here?

music/src/core/player.ts
```diff
@@ -328,9 +328,89 @@ export class SoundFontPlayer extends BasePlayer {
...
+  stop() {
+    // Do nothing.
+  }
+}
+
+/**
+ * A `NoteSequence` player based on Tone.js that allows arbitrary callbacks
+ * when playing each note.
+ */
+export class PlayerWithCallback extends Player {
```
adarob: It seems like it would be better/easier just to add the callback as an option to Player.

music/demos/callback_player.html
```diff
@@ -0,0 +1,83 @@
+<html>
```
adarob: you need to add callback_player.ts to build_demos.sh and a link to this page from the index.html

adarob: you need to add callback_player.ts to build_demos.sh and add a link to callback_player.html from index.html

music/src/core/player.ts
psc-g: the way it's coded up i would have to add the constructor options to BasePlayer, since i modify the functionality of `start()`. just want to confirm you're ok with that before proceeding.

music/src/core/player.ts
psc-g: done

music/src/core/player.ts
psc-g: done

music/src/core/player.ts
psc-g: done

music/src/core/player.ts
psc-g: done

music/src/index.ts
psc-g: i had originally tried that but it didn't work, giving me the error:
`TypeScript error: callback_player.ts(99,23): Error TS2339: Property 'constants' does not exist on type 'typeof "../src/index"'.`

given that the import line is `import * as mm from '../src/index';` it makes sense that it can't find it unless i add that line.

music/src/core/player.ts
psc-g: done

music/src/core/player.ts
psc-g: this is to ensure we only play the click when selected. so the logic goes:
if this.playClick: play all notes
if !this.playClick: only play notes when != one of the click pitches

i've added abuttons in the demo page to also play without click to make sure this works.

music/demos/callback_player.html
psc-g: done

psc-g: done

music/demos/callback_player.ts
```diff
@@ -0,0 +1,213 @@
...
+        color = 'black';
+      }
+      setTimeout(() => {
+        this.keyDivs[relativePitch][i].style.background = color;
+      }, 100);
+    }
+  }
+
+  run(t: number, n: mm.NoteSequence.INote) {
+    if (n.isDrum) {
```
teropa: To get the visualization to sync up exactly with the sound, you can schedule it with [Tone.Draw](https://tonejs.github.io/docs/r12/Draw). Otherwise it'll show early, as Tone events are fired slightly before the scheduled time.

music/src/core/player.ts
adarob: Yes, that's fine as long as it works for all of the child classes.

music/src/index.ts
adarob: gotcha, and how do you find it after you add the line? is it mm.constants.*?

psc-g: no need for `constants`, just `mm.HI_CLICK_PITCH` for instance.

music/demos/callback_player.ts
psc-g: thanks! done

music/src/core/player.ts
psc-g: moved all click and callback logic into `BasePlayer`. made code a lot cleaner, and now all can benefit from this! :)

music/src/core/player.ts
```diff
@@ -39,21 +39,76 @@ function compareQuantizedNotes(a: NoteSequence.INote, b: NoteSequence.INote) {
...
 
+/**
+ * An abstract base class for providing arbitrary callbacks for each note
+ * played.
+ */
+export abstract class BasePlayerCallback {
+
+  /* Will be called for each time/note pair in a sequence being played.
+   */
+  abstract run(n: NoteSequence.INote): void;
```
adarob: should we also pass in time?

music/src/core/player.ts
```diff
@@ -39,21 +39,76 @@ function compareQuantizedNotes(a: NoteSequence.INote, b: NoteSequence.INote) {
...
-  constructor() {
+  /**
+   *   `BasePlayer` constructor.
+   *
+   *   @param playClick A boolean, determines whether the click will be played.
+   *   @param callbackObject An optional BasePlayerCallback, specifies an
+   *     object that contains run() and stop() methods to invode during
+   *     playback.
+   */
+  constructor(playClick=true, callbackObject?: BasePlayerCallback) {
```
adarob: default should be false for playclick

adarob: Perhaps it should be clickSignature instead of playClick, which may be udnefined to disable. Perhaps we only support 4/4 at first, but this will make it easier to extend later.

music/src/index.ts
adarob: we want these to be under mm.constants actually, same with mm.data

psc-g: how do you do this with typescript? i tried `export * from './core/constants' as constants;` (and a few variants) but it didn't work.
or should i just import these explicitly in callback_player?

music/src/core/player.ts
psc-g: the reason i'm not passing in time now is that time is now used in the call to `Tone.Draw`. since the whole point of this is to sync arbitrtary events with note playback, it seemed unnecessary to pass in time.
if you'd prefer though, i can obviously pass it in and leave it unused for now.

music/src/core/player.ts
psc-g: made false by default.
do NoteSequences normally have time signature defined? if so, it probably makes sense to just use that rather than allowing arbitrary time signatures that don't necessarily correspond to the note sequence.
thoughts?

music/src/core/player.ts
adarob: Ack

music/src/index.ts
adarob: Ah, I think you need to add them to ./core/index.ts. In fact, the data export is already there.

psc-g: cool, thanks! done.

PR #104, opened 2018-06-28
Add new 4-bar melody model.

adarob:


adarob:
@teropa, this new checkpoint directory contains an `attribute_vecs.json` file with vectors going in the positive direction for each attribute.

teropa:
Cool, I'll take it for a spin!

adarob:
Just FYI, the attribute vectors don't work quite as well on this model as
in the paper since it is a shrunken version of the model to reduce the
checkpoint size.

teropa:
The checkpoint at https://storage.googleapis.com/magentadata/js/checkpoints/music_vae/mel_4bar_small_q2 seems to be missing a `config.json`, causing magenta.js to not want to load it

adarob:
Try now!

teropa:
Yesss! Thanks

PR #110, opened 2018-07-24
Tempo switch

psc-g:
Add functionality to switch tempo during playback.


music/src/core/player.ts
```diff
@@ -119,10 +134,12 @@ export abstract class BasePlayer {
    * @returns a Promise that resolves when playback is complete.
    */
   start(seq: INoteSequence, qpm?: number): Promise<void> {
-    if (this.playClick) {
+    let isQuantized = sequences.isQuantizedSequence(seq);
+    if (this.playClick && isQuantized) {
       seq = this.makeClickSequence(seq);
     }
-    if (sequences.isQuantizedSequence(seq)) {
+    Tone.Transport.bpm.value = 120;
```
adarob: Why 120 here? Shouldn't it be set to the QPM of the unquantized sequence?

music/src/core/player.ts
```diff
@@ -78,13 +79,27 @@ export abstract class BasePlayer {
    *   @param callbackObject An optional BasePlayerCallback, specifies an
    *     object that contains run() and stop() methods to invode during
    *     playback.
+   *   @param bpm An optional number specifying the starting BPM for playback.
    */
-  constructor(playClick = false, callbackObject?: BasePlayerCallback) {
+  constructor(playClick = false, callbackObject?: BasePlayerCallback,
+              bpm = 60) {
```
adarob: I don't think we actually need to set an initial BPM in the constructor since it is set in `start()` now.

music/src/core/player.ts
```diff
@@ -78,13 +79,27 @@ export abstract class BasePlayer {
...
-    Tone.Transport.bpm.value = 60;
+    this.desiredBPM = bpm;
+  }
+
+  /**
+   * Changes the bpm of the playback.
+   *
+   * @param bpm A number, the new bpm to use.
+   */
+  setBPM(bpm: number) {
```
adarob: We use `qpm` elsewhere, and we should be consistent. Maybe calling this `setTempo(qpm: number)` would be even more clear?

music/src/core/player.ts
```diff
@@ -140,6 +157,7 @@ export abstract class BasePlayer {
         }, t);
       }
     }, seq.notes.map(n => [n.startTime, n]));
+    Tone.Transport.bpm.value = this.desiredBPM;
```
adarob: If a new qpm is passed in the arguments, does that become the desired qpm?

music/src/core/player.ts
psc-g: what i thought i had to do for this to work was to first select a "standard" qpm (i picked 120 because that's what's in constants.DEFAULT_QUARTERS_PER_MINUTE), enqueue the notes, and then switch to the desired qpm.

i guess i was also assuming an unquantized sequence shouldn't include qpm (the one in my example doesn't). do unquantized sequences also normally specify qpm in tempos[0].qpm?

music/src/core/player.ts
psc-g: but it's only allowed for quantized sequences, so for quantized sequences you could only do it via the constructor or via setBPM. i can remove the constructor one if you feel it would be simpler to just do it via setBPM, or i could also remove the restriction in start().

music/src/core/player.ts
psc-g: done.

music/src/core/player.ts
psc-g: ah good catch, fixed.

music/src/core/player.ts
```diff
@@ -78,13 +79,25 @@ export abstract class BasePlayer {
    *   @param callbackObject An optional BasePlayerCallback, specifies an
    *     object that contains run() and stop() methods to invode during
    *     playback.
+   *   @param qpm An optional number specifying the starting BPM for playback.
    */
-  constructor(playClick = false, callbackObject?: BasePlayerCallback) {
+  constructor(playClick = false, callbackObject?: BasePlayerCallback,
+              qpm = 60) {
```
adarob: I don't think we should specify a default qpm in the constructor. Instead, we should play the sequence back at the qpm in given as an argument to `start()`, the desired QPM if `setTempo` is called and no argument is given, the NoteSequence tempos[0].qpm if no argument is given and `setTempo` hasn't been called, or the DEFAULT_QUARTERS_PER_MINUTE if none of the above.

music/src/core/player.ts
adarob: You should use the tempos[0].qpm if it exists, or the default if it does not.

psc-g: done.

music/src/core/player.ts
psc-g: done.

PR #119, opened 2018-08-03
Additional fixes for new tfjs version. Corrected doc output dir for music.

adarob:


music/src/music_vae/model.ts
```diff
@@ -341,7 +341,9 @@ class BaseDecoder extends Decoder {
         if (this.nade == null) {
           const timeLabels =
               (temperature ?
-                   tf.multinomial(logits.div(tf.scalar(temperature)), 1)
+                   tf.multinomial(
```
adarob: @nsthorat @dsmilkov any idea why this change was necessary? I was getting this error:

```
src/music_vae/model.ts:345:22 - error TS2345: Argument of type 'Tensor<Rank>' is not assignable to parameter of type 'number | boolean | Int32Array | Uint8Array | Float32Array | number[] | boolean[] | number[][] | n...'.
  Type 'Tensor<Rank>' is not assignable to type 'Tensor<Rank.R2>'.

345                      logits.div(tf.scalar(temperature, 'float32')), 1).as1D() :
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

nsthorat: What's happening is tf.multinomial expects a Tensor2D (statically). With simply logits.div(temperature) you are broadcast dividing a matrix by a scalar. We cant infer that broadcast type lookup (tf.Tensor2D + scalar = tf.Tensor2D) statically, so you have to cast the output of logits.div(temperature) as Tensor2D before passing it to tf.multinomial.

The reason that logits.divStrict fixes this error is that divStrict forces the two arguments to be of the same rank, and then we can generically return a tensor of that rank. This is why typescript knows it's a Tensor2D (again, all you have to do is cast the output of the broadcast to Tensor2D above).

nsthorat: I think what changed is that tf.multinomial got stricter (Tensor => Tensor2D signature) to mimic tensorflow.

PR #123, opened 2018-08-13
add a travis config

notwaldorf:
I've added the same config from the `tjfs-core` repo. This work for now, since `music` has tests but `sketch` doesn't, but we can add a [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#build-matrix) if more subdirectories have tests.

To actually run the tests, I think Travis needs to be added to this repo, possibly by [going here](https://github.com/settings/installations), and adding Travis CI as an installed app.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: node_js
+node_js:
+  - '8'
+  - '10'
+script:
+  - cd music && yarn install
+  - npm run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
-node_js:
-  - '8'
-  - '10'
+node_js: stable
 script:
   - cd music && yarn install
   - npm run test
--- a/music/README.md
+++ b/music/README.md
@@ -8,19 +8,19 @@ For the Python TensorFlow implementations, see the [main Magenta repo](https://g
 
 ## Contents
 
-* [Example Applcations](#example-applications)
-* [Supported Models](#supported-models)
-* [Getting Started](#getting-started)
-* [Model Checkpoints](#model-checkpoints)
+- [Example Applications](#example-applications)
+- [Supported Models](#supported-models)
+- [Getting Started](#getting-started)
+- [Model Checkpoints](#model-checkpoints)
 
 ## Example Applications
 
 Here are a few applications built with MagentaMusic.js:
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: node_js
+node_js:
+  - '8'
+  - '10'
```
cghawthorne: i'm not super familiar with travis. does this mean it will run the tests 2 times on both node 8 and node 10?

At this point, I think we're fine just supporting the latest version of node. Would it be easier to just use 10? I'm not sure what the normal convention is.

notwaldorf: yeah, it'll run two sets of tests. i'll make it so that it always runs `stable`. i can also run it on multiple browsers if you want, if you think there's a chance for cross-platform problems

cghawthorne: Oh, interesting. I think browser compatibility is more a concern at this point than Node compatibility. Maybe test on the latest Chrome and Safari for both desktop and mobile? @adarob do you have thoughts?

notwaldorf: Oof, I don't think Travis supports mobile (or Windows, yet), but for now I can test on Chrome/FF on Linux, and Safari on OSX.

notwaldorf: Oh, i just realized none of the tests are browser tests -- they're just in node, so the browsers don't actually matter. 🤦‍♀️

music/README.md
```diff
@@ -8,19 +8,19 @@ For the Python TensorFlow implementations, see the [main Magenta repo](https://g
...
+- [Example Applications](#example-applications)
+- [Supported Models](#supported-models)
+- [Getting Started](#getting-started)
+- [Model Checkpoints](#model-checkpoints)
 
 ## Example Applications
 
 Here are a few applications built with MagentaMusic.js:
 
-* [Beat Blender](https://g.co/beatblender) by [Google Creative Lab](https://github.com/googlecreativelab)
```
notwaldorf: @cghawthorne I had `prettify` left on which fixed a bunch of whitespace and stuff, let me know if you want me to undo this

cghawthorne: I think it's fine.

notwaldorf:
🎉🎉🎉 
screen shot 2018-08-14 at 10 52 59 am

PR #124, opened 2018-08-14
Add new method to return the sequence of probabilities from the music rnn

notwaldorf:
Added this as a separate method to not break the compatibility of `continueSequence` (which can't really change its return type either)

In general:
download

### DIFF
--- a/music/src/music_rnn/model.ts
+++ b/music/src/music_rnn/model.ts
@@ -205,6 +205,34 @@ export class MusicRNN {
   async continueSequence(
       sequence: INoteSequence, steps: number, temperature?: number,
       chordProgression?: string[]): Promise<INoteSequence> {
+    const result = await this.continueSequenceImpl(
+        sequence, steps, temperature, chordProgression, false);
+    return result.sequence;
+  }
+
+  /**
+   * Continues a provided quantized NoteSequence containing a monophonic melody,
+   * and returns the sequence of probabilities at each step.
+   *
+   * @param sequence The sequence to continue. Must be quantized.
+   * @param steps How many steps to continue.
+   * @param temperature (Optional) The softmax temperature to use when sampling
+   * from the logits. Argmax is used if not provided.
+   * @param chordProgression (Optional) Chord progression to use as
+   * conditioning.
+   */
+  async continueSequenceAndReturnProbabilities(
+      sequence: INoteSequence, steps: number, temperature?: number,
+      chordProgression?: string[]):
+      Promise<{sequence: Promise<INoteSequence>; probs: tf.Tensor[]}> {
+    return this.continueSequenceImpl(
+        sequence, steps, temperature, chordProgression, true);
+  }
+
+  private async continueSequenceImpl(
+      sequence: INoteSequence, steps: number, temperature?: number,
+      chordProgression?: string[], computeProbs?: boolean):
+      Promise<{sequence: Promise<INoteSequence>; probs: tf.Tensor[]}> {
     sequences.assertIsRelativeQuantizedSequence(sequence);
 
     if (this.chordEncoder && !chordProgression) {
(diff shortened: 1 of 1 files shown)


music/src/music_rnn/model.ts
```diff
@@ -205,6 +205,34 @@ export class MusicRNN {
...
+   * @param steps How many steps to continue.
+   * @param temperature (Optional) The softmax temperature to use when sampling
+   * from the logits. Argmax is used if not provided.
+   * @param chordProgression (Optional) Chord progression to use as
+   * conditioning.
+   */
+  async continueSequenceAndReturnProbabilities(
+      sequence: INoteSequence, steps: number, temperature?: number,
+      chordProgression?: string[]):
+      Promise<{sequence: Promise<INoteSequence>; probs: tf.Tensor[]}> {
```
adarob: We should return the probs as a FloatArray[] or number[][] or whatever.

music/src/music_rnn/model.ts
```diff
@@ -263,6 +296,7 @@ export class MusicRNN {
     // Initialize with input.
     inputs = inputs.toFloat();
     const samples: tf.Tensor1D[] = [];
+    const probs: tf.Tensor[] = [];
```
adarob: tf.Tensor2D[]

music/src/music_rnn/model.ts
```diff
@@ -275,17 +309,29 @@ export class MusicRNN {
...
+        let sampledOutput;
+        if (computeProbs) {
+          // We are also returning the sequence of probabilities, so calculate
+          // those before sampling.
+          const theseProbs = temperature ?
+              tf.softmax(logits.div(tf.scalar(temperature))) :
+              tf.softmax(logits);
+          probs.push(theseProbs);
+          sampledOutput =
+              tf.multinomial(theseProbs as tf.Tensor2D, 1, null, true).as1D();
```
adarob: use `undefined` instead of `null`

music/src/music_rnn/model.ts
```diff
@@ -275,17 +309,29 @@ export class MusicRNN {
...
-                 tf.multinomial(
-                       logits.div(tf.scalar(temperature)) as tf.Tensor2D, 1)
-                     .as1D() :
-                 logits.argMax(1).as1D());
+
+        let sampledOutput;
+        if (computeProbs) {
+          // We are also returning the sequence of probabilities, so calculate
+          // those before sampling.
+          const theseProbs = temperature ?
```
adarob: For simplicity, you can just compute the probs in both cases and use it in the multinomial. The only time you shouldn't compute the probs is if computeProbs is false and temperature is not defined.

music/src/music_rnn/model.ts
```diff
@@ -232,20 +260,25 @@ export class MusicRNN {
...
-    return result;
+        samplesAndProbs.samples, sequence.quantizationInfo.stepsPerQuarter);
+    oh.samples.dispose();
+    return {sequence: result, probs: samplesAndProbs.probs};
   }
 
   private sampleRnn(
       inputs: tf.Tensor2D, steps: number, temperature: number,
-      controls?: tf.Tensor2D, auxInputs?: tf.Tensor2D) {
+      controls?: tf.Tensor2D, auxInputs?: tf.Tensor2D, computeProbs?: boolean) {
```
adarob: returnProbs?

music/src/music_rnn/model.ts
```diff
@@ -193,7 +193,7 @@ export class MusicRNN {
   }
 
   /**
-   * Continues a provided quantized NoteSequence containing a monophonic melody.
+   * Continues a provided quantized NoteSequence containing a melody.
```
adarob: also don't mention melody:)

just "...quantized NoteSequence."

music/src/music_rnn/model.ts
```diff
@@ -263,6 +306,7 @@ export class MusicRNN {
     // Initialize with input.
     inputs = inputs.toFloat();
     const samples: tf.Tensor1D[] = [];
+    const probs: tf.Tensor2D[] = [];
```
adarob: actually this should be tf.Tensor1D[] since our batch size is always 1, sorry about that

PR #126, opened 2018-08-18
Add a midi visualizer

notwaldorf:
- made `mm.Visualizer` in `core`, which makes something look like this:
screen shot 2018-08-17 at 4 58 01 pm

- in the demo, I connected it to a Player to make it play in sync. I had originally made this as a web component, but loading web components is a bit icky with `es5`, so I passed on it.

There's a `TODO` in the code to make the repainting slightly better and more optimal, but I wanted to see first if this was barking up the right tree before I went down the road of fixing that.

Also: the demo is ugly, but I am planning a followup PR to make all the demos a bit prettier (and easier to understand their input/output)

I uhhhh don't know how to test this, or if it's worth testing. 😭

PR #128, opened 2018-08-21
Prettify demos

notwaldorf:
Only look at this PR after #126 is merged.

This looks scary and giant, but it's very similar changes across all demos
- added a little blurb at the top to explain what the demo does.
- used the `mm.Visualizer` for each of the NoteSequences:
screen shot 2018-08-20 at 6 26 54 pm
- the NoteSequences are always initially collapsed, because for the VAE demos there's like a billion of them and it looked a bit intense, but I can make it so that only the text of the note sequence is collapsed. The way it is now:
screen shot 2018-08-20 at 6 29 04 pm

The Multitrack demo is the only one that _doesn't_ have visualizations because bad things (can't connect to the audio node) happen if I add a `callbackObject` to a SoundfontPlayer and I haven't figured out why that is yet.


music/demos/multitrack.html
```diff
@@ -3,7 +3,7 @@
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-    http://www.apache.org/licenses/LICENSE-2.0
+    http:</b>//www.apache.org/licenses/LICENSE-2.0
```
adarob: u w0t m8?

notwaldorf: it's a bold license, adam. bold.

PR #133, opened 2018-08-23
add a mm.Recorder

notwaldorf:
- added a demo for both one-off recording, and streaming recording.
- I haven't connected this to a tempo, so the sequence returned is always non-quantized (for now)

### DIFF
--- a/music/src/core/recorder.ts
+++ b/music/src/core/recorder.ts
@@ -81,7 +81,8 @@ export class Recorder {
   }
 
   /**
-   * Returns a list of all the currently connected MIDI inputs.
+   * Returns a list of all the MIDI inputs that are currently available to
+   * record.
    */
   getMIDIInputs(): WebMidi.MIDIInput[] {
     return this.midiInputs;
--- a/music/src/core/recorder.ts
+++ b/music/src/core/recorder.ts
@@ -31,25 +31,26 @@ export abstract class BaseRecorderCallback {
 }
 
 /**
- * Records
+ * Class that records MIDI from any MIDI connected instrument, and converts it
+ * to a `NoteSequence`.
  */
 export class Recorder {
-  private _isRecording: boolean;
+  private recording: boolean;
   public callbackObject: BaseRecorderCallback;
   private firstNoteTimestamp: number;
   private notes: NoteSequence.Note[] = [];
   private downNotes: {[pitch: number]: NoteSequence.Note} = {};
 
   /**
-   *    `Recorder` constructor.
+   * `Recorder` constructor.
    *
-   *     @param callbackObject An optional BasePlayerCallback, specifies an
-   *     object that contains run() and stop() methods to invode during
-   *     playback.
+   * @param callbackObject An optional BasePlayerCallback, specifies an
+   * object that contains run() and stop() methods to invode during
+   * playback.
    */
   constructor(callbackObject?: BaseRecorderCallback) {
     this.callbackObject = callbackObject;
-    this._isRecording = false;
+    this.recording = false;
   }
 
   /**
@@ -58,7 +59,7 @@ export class Recorder {
    * they will be ignored until `startRecording()` is called.
    */
   async initialize() {
-    // Start up WebMidi
+    // Start up WebMidi.
     await (navigator as Navigator)
         .requestMIDIAccess()
         .then(
@@ -67,6 +68,7 @@ export class Recorder {
   }
 
   private midiReady(midi: WebMidi.MIDIAccess) {
+    console.log('Initialized Recorder');
     const inputs = midi.inputs.values();
     for (let input = inputs.next(); input && !input.done;
          input = inputs.next()) {
@@ -78,14 +80,14 @@ export class Recorder {
   }
 
   isRecording() {
-    return this._isRecording;
+    return this.recording;
   }
 
   /**
    * Starts listening to MIDI events and records any messages received.
    */
   startRecording() {
-    this._isRecording = true;
+    this.recording = true;
 
     // Reset all the things needed for the recording.
     this.firstNoteTimestamp = undefined;
@@ -100,7 +102,7 @@ export class Recorder {
    * @returns a `NoteSequence` containing all the recorded notes.
    */
   stopRecording(): NoteSequence {
-    this._isRecording = false;
+    this.recording = false;
     if (this.notes.length === 0) {
       return null;
     }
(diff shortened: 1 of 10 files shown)


music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * An abstract base class for providing arbitrary callbacks for each note
+ * recorded.
+ */
+export abstract class BaseRecorderCallback {
+  /**
+   * Will be called for each time a note is recorded.
+   *
+   * @param seq The note sequence up to this point
```
adarob: .

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * Class that records MIDI from any MIDI connected instrument, and converts it
+ * to a `NoteSequence`.
+ */
+export class Recorder {
+  private recording: boolean;
+  public callbackObject: BaseRecorderCallback;
+  private firstNoteTimestamp: number;
+  private notes: NoteSequence.Note[] = [];
+  private downNotes: {[pitch: number]: NoteSequence.Note} = {};
```
adarob: onNotes I think is better -- also how about using Map<number, NoteSequence.Note>?

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+    this.callbackObject = callbackObject;
+    this.recording = false;
+  }
+
+  /**
+   * Initializes the `Recorder` by requesting MIDI access from the browser,
+   * and connecting to any available MIDI inputs. If MIDI events are emitted,
+   * they will be ignored until `startRecording()` is called.
+   */
+  async initialize() {
```
adarob: you should be able to list available MIDI inputs (separate method) and optionally select a subset of them here

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  isRecording() {
+    return this.recording;
+  }
+
+  /**
+   * Starts listening to MIDI events and records any messages received.
+   */
+  startRecording() {
```
adarob: start / stop for consistency with the player

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Stops listening to MIDI events. Note that all the MIDI inputs will still
+   * be connected after calling this method, but any messages received from them
+   * will be ignored.
+   * @returns a `NoteSequence` containing all the recorded notes.
+   */
+  stopRecording(): NoteSequence {
+    this.recording = false;
```
adarob: end any open notes

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+    // MIDI commands we care about. See
+    // tslint:disable-next-line
+    // http://webaudio.github.io/web-midi-api/#a-simple-monophonic-sine-wave-midi-synthesizer.
+    const NOTE_DOWN = 9;
+    const NOTE_UP = 8;
+
+    const cmd = event.data[0] >> 4;
+    const pitch = event.data[1];
+    const velocity = (event.data.length > 2) ? event.data[2] : 1;
+    if (cmd === NOTE_DOWN) {
```
adarob: NOTE_ON

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+      const note = new NoteSequence.Note();
+      note.pitch = pitch;
+      note.startTime = (event.timeStamp - this.firstNoteTimestamp) / 1000;
+      note.velocity = velocity;
+      // Save this note so that we can finish it when we receive the note up
+      this.downNotes[pitch] = note;
+    } else if (cmd === NOTE_UP) {
```
adarob: NOTE_OFF

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+      // Find the note that was originall pressed to finish it.
```
adarob: originally

music/src/core/recorder.ts
```diff
@@ -0,0 +1,168 @@
...
+      const note = this.downNotes[pitch];
+      if (note) {
+        // Notes are saved in seconds, timestamps are in milliseconds.
```
adarob: same comment above or just make 1000 be a constant : MILLIS_PER_SECOND

music/src/core/recorder.ts
notwaldorf: oh yeah that's definitely what I was trying to say

music/src/core/recorder.ts
notwaldorf: I added a `getMIDIInputs` and an argument to `start()` but I'm not sure that's what you meant.

music/src/core/recorder.ts
```diff
@@ -72,27 +72,47 @@ export class Recorder {
...
+   * Returns a list of all the currently connected MIDI inputs.
+   */
+  getMIDIInputs(): WebMidi.MIDIInput[] {
+    return this.midiInputs;
+  }
+
   /**
    * Starts listening to MIDI events and records any messages received.
+   *
+   * @param input An optional MIDIInput, that specifies which input
```
adarob: allow a list?

music/src/core/recorder.ts
```diff
@@ -72,27 +72,47 @@ export class Recorder {
...
+      this.midiInputs.push(input.value);
     }
   }
 
   isRecording() {
     return this.recording;
   }
 
+  /**
+   * Returns a list of all the currently connected MIDI inputs.
```
adarob: Are these the ones we are recording? We also need a (static?) method to tell us what inputs are available so that users can choose the one they want. Or is this just easy enough for them to do themselves with WebMIDI?

notwaldorf: No, these are all of the ones that are connected (i.e. not recording necessarily). I think this is the list you're talking about. (If inputs aren't "connected" they're not seen by the MidiAccess interface)

notwaldorf: so you could do like `recorder.start(recorder.getMIDIInputs[0])`

music/src/core/recorder.ts
```diff
@@ -0,0 +1,212 @@
...
+      this.midiInputs.push(input.value);
+    }
+  }
+
+  isRecording() {
+    return this.recording;
+  }
+
+  /**
+   * Returns a list of all the currently connected MIDI inputs.
```
adarob: ...all the MIDI inputs that are currently available to record.

PR #135, opened 2018-08-24
Add checkpoint downloader script.

### LINKED ISSUE #127: How can I download the pre-trained models?
I would like to download all multi-track pre-trained models, but when I clicked one of the links, it displayed the following error message:
```
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Details>
No such object: magentadata/js/checkpoints/music_vae/multitrack_med_chords_fb256
</Details>
</Error>
```

adarob:
Closes #127.


scripts/checkpoint_downloader.py
```diff
@@ -0,0 +1,88 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""
+Download all necessary files from a Magenta.js checkpoint URL.
+
+Example usage:
+
+$ python3 checkpoint_downloader /checkpoint/url/ /path/to/output
```
cghawthorne: Any reason for the python3 vs python mismatch between this and the script?

scripts/checkpoint_converter.py
```diff
@@ -22,7 +22,7 @@
 
 Example usage:
 
-python3 checkpoint_converter /path/to/checkpoint.ckpt /path/to/output \
+$ python checkpoint_converter /path/to/checkpoint.ckpt /path/to/output \
```
notwaldorf: why is this `python` but the instructions in the downloader say `python3 checkpoint_downloader /checkpoint/url/ /path/to/output`? I know 0 python

adarob: oversight

scripts/checkpoint_downloader.py
adarob: oversight

PR #138, opened 2018-08-25
Fix final discrepancies against Python Onsets & Frames code.

adarob:


music/demos/transcription.ts
```diff
@@ -47,6 +74,11 @@ async function transcribe() {
   writeTimer('transcription-time', start);
   writeNoteSeqs('transcription-results', [ns], undefined, true);
   oaf.dispose();
+
+  document.getElementById('ns-match').innerHTML =
+      notesMatch(ns.notes, expectedNs.notes) ?
+      '<font color="green">TRUE</font>' :
```
notwaldorf: Oh no, we don't use `<font>` anymore. How about:
`<span style="color:green>TRUE</span>` (and the same for false)

music/src/core/data.ts
```diff
@@ -24,7 +24,6 @@
  * Imports
  */
 import * as tf from '@tensorflow/tfjs-core';
-import {isNullOrUndefined} from 'util';
```
notwaldorf: 🙌

music/demos/transcription.ts
```diff
@@ -31,6 +31,33 @@ const EXPECTED_NS_URL = `${
     TRANS_CKPT_DIR}/onsets_frames_htk0/MAPS_MUS-mz_331_3_ENSTDkCl.melhtk0-250frames.ns.json`;
 // tslint:enable:max-line-length
 
+function sortNotes(notes: mm.NoteSequence.INote[]) {
```
notwaldorf: How do you feel about moving this to `common.ts`? I feel it's a useful one.

music/demos/transcription.ts
```diff
@@ -31,6 +31,33 @@ const EXPECTED_NS_URL = `${
...
+      return -1;
+    }
+    if (a.endTime < b.endTime) {
+      return -1
+    }
+    return a.pitch - b.pitch;
+  });
+}
+
+function notesMatch(a: mm.NoteSequence.INote[], b: mm.NoteSequence.INote[]) {
```
notwaldorf: Alternatively, this could also be a method on `NoteSequence` itself. Something like `matches`?

notwaldorf: (you don't have to do this, it's just a suggestion)

music/src/core/data.ts
```diff
@@ -191,7 +190,7 @@ export class DrumsConverter extends DataConverter {
 
   constructor(args: DrumsConverterArgs) {
     super(args);
-    this.pitchClasses = isNullOrUndefined(args.pitchClasses) ?
+    this.pitchClasses = args.pitchClasses == undefined  ?
```
notwaldorf: `===`

PR #142, opened 2018-08-29
Remove used of normalized option in tf.multinomial since it is not supported by tfjs-node.

adarob:
fyi, @tambien

### DIFF
--- a/music/src/core/data.ts
+++ b/music/src/core/data.ts
@@ -190,10 +190,7 @@ export class DrumsConverter extends DataConverter {
 
   constructor(args: DrumsConverterArgs) {
     super(args);
-    this.pitchClasses =
-        (args.pitchClasses === null || args.pitchClasses === undefined) ?
-        DEFAULT_DRUM_PITCH_CLASSES :
-        args.pitchClasses;
+    this.pitchClasses = args.pitchClasses || DEFAULT_DRUM_PITCH_CLASSES;
     this.pitchToClass = new Map<number, number>();
     for (let c = 0; c < this.pitchClasses.length; ++c) {  // class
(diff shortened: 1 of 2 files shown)


music/src/core/data.ts
```diff
@@ -190,7 +190,8 @@ export class DrumsConverter extends DataConverter {
 
   constructor(args: DrumsConverterArgs) {
     super(args);
-    this.pitchClasses = args.pitchClasses === undefined  ?
+    this.pitchClasses =
+        (args.pitchClasses === null || args.pitchClasses === undefined) ?
```
notwaldorf: Does just `!args.pitchClasses ? ` work here?

music/src/music_rnn/model.ts
```diff
@@ -317,21 +317,20 @@ export class MusicRNN {
...
-                  .as1D();
+        let logits = lastOutput.matMul(this.lstmFcW).add(this.lstmFcB).as1D();
+
+        let sampledOutput: tf.Tensor1D;
+        if (temperature) {
+          logits = logits.div(tf.scalar(temperature));
+          sampledOutput = tf.multinomial(logits, 1).as1D();
         } else {
-          sampledOutput = logits.argMax(1).as1D();
+          sampledOutput = logits.argMax().as1D();
```
notwaldorf: wait, we don't want softmax here anymore?

music/src/core/data.ts
adarob: that's too easy

music/src/music_rnn/model.ts
adarob: softmax doesn't change the relative magnitudes so we don't need it before an argmax

PR #143, opened 2018-08-29
Remove more isNullOrUndefined uses and bump version.

adarob:


notwaldorf:
I think this should be a minor bump — 1.2.0. There’s a bunch of new features added and that’s what semver says to do. :)

notwaldorf:
But there’s also a visualizer and a recorder. Hmmm maybe you shouldn’t do a release then?

notwaldorf:
(release numbers tell people what to expect. This release adds new things, rather than fix existing ones, so it’s a bit weird to mark it as such imo. But I won’t block it or anything, I’m just uhhh offering an opinion)

adarob:
This release is just to fix these bugs and unblock Yotam with the node-tf
stuff.

notwaldorf:
Also I think most of these as long as they’re not numbers could be “if !foo” but I think this is the way safer bet. Anyway, ship it squirrel! :shipit:

PR #144, opened 2018-08-30
fix transcription visualizer raciness

notwaldorf:
Soooo the visualizers in `transcription.html` were jumping because `notesMatch` would re-sort the NoteSequence after the Visualizer was initialized, so at some racy point it would re-draw to the new one. Now `notesMatch` makes a copy and errrthing is fine.

Also:
- did a small run through of `tslint` fixes
- added this way better NoteSequence to the Visualizer demo because it's better than Twinkle Twinkle.

### DIFF
--- a/music/demos/visualizer.ts
+++ b/music/demos/visualizer.ts
@@ -16,9 +16,14 @@
  */
 
 import * as mm from '../src/index';
-import {FULL_TWINKLE} from './common';
+
+import {CHECKPOINTS_DIR} from './common';
 
 const MIDI_URL = './melody.mid';
+// tslint:disable:max-line-length
+const EXPECTED_NS_URL = `${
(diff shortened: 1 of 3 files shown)


music/demos/visualizer.ts
```diff
@@ -47,10 +52,15 @@ const canvas = document.getElementById('canvas') as HTMLCanvasElement;
 // Set up some event listeners
 urlBtn.addEventListener('click', () => fetchMidi(MIDI_URL));
 playBtn.addEventListener('click', () => startOrStop());
-seqBtn.addEventListener('click', () => initPlayerAndVisualizer(FULL_TWINKLE));
+seqBtn.addEventListener('click', () => {
+  fetch(EXPECTED_NS_URL).then((response) => response.json()).then((ns => {
+    console.log(ns);
```
adarob: nolog?

PR #147, opened 2018-08-30
Handle polyphony in MelodyConverter by keeping the highest note at any start step.

adarob:
This mimics the behavior of the Magenta Python library.

### DIFF
--- a/music/src/core/data.ts
+++ b/music/src/core/data.ts
@@ -383,7 +383,6 @@ export class MelodyConverter extends DataConverter {
         });
     const mel = tf.buffer([numSteps], 'int32');
     let lastStart = -1;
-    console.log(sortedNotes);
     sortedNotes.forEach(n => {
       if (n.quantizedStartStep === lastStart) {
         if (this.disallowPolyphony) {
--- a/music/src/core/data.ts
+++ b/music/src/core/data.ts
@@ -389,6 +389,9 @@ export class MelodyConverter extends DataConverter {
         if (this.disallowPolyphony) {
           throw new Error('`NoteSequence` is not monophonic.');
         } else {
+          // Keep highest note.
+          // Notes are sorted by pitch descending, so if a note is already at
+          // this position its the highest pitch.
           return;
         }
       }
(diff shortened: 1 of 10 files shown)


adarob:
fyi @tamblien

music/src/core/data.ts
```diff
@@ -361,19 +366,34 @@ export class MelodyConverter extends DataConverter {
...
+        noteSequence.notes.sort((n1, n2) => {
+          if (n1.quantizedStartStep === n2.quantizedStartStep) {
+            return n2.pitch - n1.pitch;
+          }
+          return n1.quantizedStartStep - n2.quantizedStartStep;
+        });
     const mel = tf.buffer([numSteps], 'int32');
-    let lastEnd = -1;
+    let lastStart = -1;
+    console.log(sortedNotes);
```
cghawthorne: should remove

music/src/core/data.ts
```diff
@@ -361,19 +366,33 @@ export class MelodyConverter extends DataConverter {
     super(args);
     this.minPitch = args.minPitch;
     this.maxPitch = args.maxPitch;
+    this.disallowPolyphony = args.disallowPolyphony;
```
notwaldorf: should this have a default of true somehow?

adarob: I'm actually defaulting it to false since I believe that is more in line with the expected behavior. @tambien can maybe speak to this as well?

adarob: after discussion, changed to `ignorePolyphony` with default value of true

music/src/core/data.ts
```diff
@@ -339,16 +339,21 @@ export class DrumsOneHotConverter extends DrumsConverter {
  * cause an errot to be thrown.
  * @param maxPitch The maximum pitch to model. Those above this value will
  * cause an error to be thrown.
+ * @param ignorePolpyhony (default: trie) If false, an error will be raised
```
notwaldorf: typo: "trie" (twice)

PR #151, opened 2018-09-06
Use velocity in drum/synth players

notwaldorf:
Yay! Tone.js already had a thing for this!

### DIFF
--- a/music/demos/common.ts
+++ b/music/demos/common.ts
@@ -160,6 +160,56 @@ export const MEL_TWINKLE: mm.INoteSequence = {
   totalQuantizedSteps: 32,
 };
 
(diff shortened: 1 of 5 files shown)


notwaldorf:
@adarob did you hear the clipping though?

adarob:
Where specifically should I listen for clipping?

notwaldorf:
In the last demo (that has each of the drums at a small and big velocity) one of the “loud” drums to me sounds clipped (as in it sorta gets blown out and you don’t hear the drum at all. Or the snare. I don’t know what’s it’s called)

notwaldorf:
(update: the drum doesn't get clipped because of the velocity, but because that drum (46) can't be played in succession quickly)

PR #158, opened 2018-09-09
Lots of performance improvements and a bit of refactor for O&F.

adarob:


music/src/transcription/model.ts
```diff
@@ -207,114 +196,41 @@ export class OnsetsAndFrames {
...
+      this.velocityCnn.setWeights(vars, 'velocity', 'onset_velocities');
     });
   }
 
-  /**
-   * Returns an acoustic stack without setting variables.
-   */
-  private getAcousticCnn() {
-    const acousticCnn = tf.sequential();
+  private logMessage(msg: string) {
```
notwaldorf: If you’re adding a stand-alone function, then think you should have some sort of flag to turn off debugging and wrap all the logs maybe in a if (verbose). I don’t think a lot of people like a super verbose library

music/src/transcription/model.ts
```diff
@@ -328,31 +244,216 @@ export class OnsetsAndFrames {
...
+      ]);
+    }
+    this.nn.setWeights(weights);
+  }
+}
+
+/**
+ * Helper for a Bidirecitonal LSTM layer.
+ *
+ * Implments processing the input in chunks, which is significantly more
```
notwaldorf: Typos: bidirectional, implements, caching.

music/src/transcription/model.ts
```diff
@@ -328,31 +244,216 @@ export class OnsetsAndFrames {
...
+
+    const outputFwChunks: tf.Tensor3D[] = [];
+    const outputBwChunks: tf.Tensor3D[] = [];
+    let fwState: [tf.Tensor2D, tf.Tensor2D] =
+        [tf.zeros([1, this.units]), tf.zeros([1, this.units])];
+    let bwState: [tf.Tensor2D, tf.Tensor2D] =
+        [tf.zeros([1, this.units]), tf.zeros([1, this.units])];
+    for (let i = 0; i < inputChunks.length; ++i) {
+      let input = [inputChunks[i], fwState[0], fwState[1]];
+      let res =
```
notwaldorf: Does “res” stand for result? Can we use result then? 🙏

notwaldorf: Or even better something like “prediction” or what the result actually is

music/src/transcription/model.ts
adarob: It's not really that verbose...also can't they filter based on .info?

cghawthorne: Should we have some sort of shared logging library if we're going to do stuff like that? Is there a standard one we should be using?

adarob: #159

adarob: Apparently console.info === console.log. But people can filter us completely if they want. 
If we were to have some setting, would we just have a global variable mm.verbose === true/false?

adarob: See #160.

PR #166, opened 2018-09-13
Record audio for transcription

notwaldorf:
- add a microphone -> transcription demo
- also uses the MediaRecorder polyfill so that it works on Safari
- also added spaces around the logging prefix so that it's....more spaced out

### DIFF
--- a/music/demos/transcription.html
+++ b/music/demos/transcription.html
@@ -47,6 +47,20 @@ limitations under the License.
     <p><b>Total Leaked Memory:</b> <code id="file-leaked-memory"></code></b></p>
   </section>
 
+  <h2>From a microphone recording</h2>
+  <p>You can record piano audio from a microphoe and transcribe it. If you're recording
+    something other than piano (like your voice), this will be transcribed but it
+    will probably be fairly noisy and incorrect.
+  </p>
+  <section>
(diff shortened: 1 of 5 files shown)


music/demos/transcription.html
```diff
@@ -47,6 +47,20 @@ <h2>From an uploaded audio file</h2>
     <p><b>Total Leaked Memory:</b> <code id="file-leaked-memory"></code></b></p>
   </section>
 
+  <h2>From a microphone recording</h2>
+  <p>You can record piano audio from a microphoe and transcribe it. If you're recording
```
adarob: s/microphoe/microphone/g

music/demos/transcription.ts
```diff
@@ -15,13 +15,16 @@
  * limitations under the License.
  */
 import * as tf from '@tensorflow/tfjs-core';
+//@ts-ignore
```
adarob: install types/mediarecorder and avoid all of thisany nonsense?

notwaldorf: ugh this bit of nonsense isn't cause of the types, it's cause the export is bad (it doesn't have a named export

but also i couldn't find types

Issue #130, opened 2018-08-23
Add mm.Recorder to record MIDI input with WebMIDI.


adarob:

notwaldorf:
Fixed in https://github.com/tensorflow/magenta-js/pull/133

adarob:
Reopening to also add recording w/ clicks.

PR #181, opened 2018-10-12
Add an optional click track to mm.Recorder

notwaldorf:
Fixes https://github.com/tensorflow/magenta-js/issues/130

### DIFF
--- a/music/src/core/recorder.ts
+++ b/music/src/core/recorder.ts
@@ -158,6 +158,8 @@ export class Recorder {
   private initClickLoop() {
     let clickStep = 0;
     this.clickLoop = new Tone.Loop((_:number) => {
+      // TODO(notwaldorf): It would be nice if this took into account a
+      // time signature.
       if (clickStep % 4 === 0) {
         this.loClick.triggerAttack('G5');
       } else {
--- a/music/src/core/recorder.ts
+++ b/music/src/core/recorder.ts
@@ -18,6 +18,21 @@
  * limitations under the License.
  */
 import {NoteSequence} from '../protobuf';
+import * as Tone from 'tone';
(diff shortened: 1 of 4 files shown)


notwaldorf:
Hey @tambien does this look ok re: how I am using `Tone.Loop`? Not super familiar with `Tone.js` so I want to make sure I'm not being a dummy and there's a better way

music/src/core/recorder.ts
```diff
@@ -82,6 +122,55 @@ export class Recorder {
...
+   * @param countIn True if it should play a count in.
+   */
+  enablePlayCountIn(countIn: boolean) {
+    this.config.playCountIn = countIn;
+  }
+
+  private initClickLoop() {
+    let clickStep = 0;
+    this.clickLoop = new Tone.Loop((_:number) => {
+      if (clickStep % 4 === 0) {
```
adarob: Add TODO about time sig.

music/src/core/recorder.ts
```diff
@@ -82,6 +122,57 @@ export class Recorder {
...
+    this.config.playCountIn = countIn;
+  }
+
+  private initClickLoop() {
+    let clickStep = 0;
+    this.clickLoop = new Tone.Loop((_:number) => {
+      // TODO(notwaldorf): It would be nice if this took into account a
+      // time signature.
+      if (clickStep % 4 === 0) {
+        this.loClick.triggerAttack('G5');
```
tambien: Using the passed in number parameter (_:number)  as the second argument to triggerAttack will give you much more accurate timing.

notwaldorf: I'm not sure I understand how to get "every 4 beats" from that, though

Issue #186, opened 2018-10-23
transcription/model.d.ts in npm package invalid


nehz:
Doing a `npm install @magenta/music` gets the following `node_modules/@magenta/music/es5/transcription/model.d.ts`

```
declare class OnsetsAndFrames {
    private checkpointURL;
    chunkLength: number;
    private initialized;
    private onsetsCnn;
    private onsetsRnn;
    private activationCnn;
    private frameRnn;
    private velocityCnn;
    constructor(checkpointURL: string, chunkLength?: number);
    initialize(): Promise<void>;
    dispose(): void;
    isInitialized(): boolean;
    transcribeFromMelSpec(melSpec: number[][], parallelBatches?: number): Promise<import("../../../../../../../Users/noms/Code/magenta-js/music/src/protobuf/proto").tensorflow.magenta.NoteSequence>;
    transcribeFromAudioBuffer(audioBuffer: AudioBuffer, batchSize?: number): Promise<import("../../../../../../../Users/noms/Code/magenta-js/music/src/protobuf/proto").tensorflow.magenta.NoteSequence>;
    transcribeFromAudioFile(blob: Blob): Promise<import("../../../../../../../Users/noms/Code/magenta-js/music/src/protobuf/proto").tensorflow.magenta.NoteSequence>;
    transcribeFromAudioURL(url: string): Promise<import("../../../../../../../Users/noms/Code/magenta-js/music/src/protobuf/proto").tensorflow.magenta.NoteSequence>;
    private processBatches;
    private build;
}
export { OnsetsAndFrames };
```

The paths seem to be local dev paths
```
14:90 Cannot find module '../../../../../../../Users/noms/Code/magenta-js/music/src/protobuf/proto'.
```

notwaldorf:
Oh that doesn't look good, does it? Taking a look

notwaldorf:
I think my username is there because I've done the last release, but I'm not sure why it's being resolved to that, since that's not even the signature of the function: https://github.com/tensorflow/magenta-js/blob/master/music/src/transcription/model.ts#L165. 

@adarob do you think it's because we don't have explicit `Promise<NoteSequence>` return types on those functions? lemme try and see if that fixes it

notwaldorf:
oh yeah that's totally it! sending a PR now.

notwaldorf:
Fixed in #187, i don't know why auto closing doesn't work anymore

PR #187, opened 2018-10-26
Add explicit type to transcribeFromMelSpec

notwaldorf:
Fies #186.

After this change, the types look right in the build:
screen shot 2018-10-26 at 10 47 32 am


adarob:
Closes #168

PR #192, opened 2018-11-06
fix bug in check for valid chord symbol

iansimon:


music/src/core/chords_test.ts
```diff
@@ -46,6 +49,8 @@ test('Test Chord Root', (t: test.Test) => {
   t.equal(ChordSymbols.root('Abmaj7'), 8);
   t.equal(ChordSymbols.root('D##5(add6)'), 4);
   t.equal(ChordSymbols.root('F(b7)(#9)(b13)'), 5);
+  t.throws(() => ChordSymbols.root(''), ChordSymbolException);
+  t.throws(() => ChordSymbols.root('o'), ChordSymbolException);
```
notwaldorf: I would also add a test for our friend "M" here

iansimon: Done.

PR #194, opened 2018-11-20
add a pause/resume functionality to the player

notwaldorf:
finally gave in and looked it up. turns out it was an easy fix 🤦‍♀️ 

- added a method to expose the transport state
- stop the transport on `stop()` which I don't think we were doing before

### DIFF
--- a/music/src/core/player.ts
+++ b/music/src/core/player.ts
@@ -218,8 +218,22 @@ export abstract class BasePlayer {
     Tone.Transport.start();
   }
 
+  /**
+   * Returns true iff the player is completely stopped. This will only be
+   * false after calling stop(), and will be true after calling
+   * start(), pause() or unpause().
+   */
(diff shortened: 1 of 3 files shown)


notwaldorf:
@adarob added an `isPaused`, ptal?

music/src/core/player.ts
```diff
@@ -202,9 +202,39 @@ export abstract class BasePlayer {
     this.desiredQPM = undefined;
   }
 
+  /**
+   * Pause playing the currently playing sequence right away. Call unpause()
+   * to resume
```
adarob:
```suggestion
   * to resume.
```

notwaldorf:
omg or, and hear me out: we could expose the tone.transport state? it's started/stopped/paused. Maybe that's the best instead of these weird 2 functions?

PR #197, opened 2018-11-21
Adds ability to restrict Piano Genie outputs to a particular subset of notes

chrisdonahue:
Supports use cases such as restricting to a particular scale or range of the keyboard.

### DIFF
--- a/music/src/piano_genie/model.ts
+++ b/music/src/piano_genie/model.ts
@@ -191,6 +191,7 @@ class PianoGenie {
    * @param button Button number (one of {0, 1, 2, 3, 4, 5, 6, 7}).
    * @param temperature Sampling temperature [0, 1].
    * @param seed Random seed for sampling.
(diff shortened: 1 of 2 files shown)


music/src/piano_genie/model.ts
```diff
@@ -191,6 +191,7 @@ class PianoGenie {
    * @param button Button number (one of {0, 1, 2, 3, 4, 5, 6, 7}).
    * @param temperature Sampling temperature [0, 1].
    * @param seed Random seed for sampling.
+   * @param keySubset Restrict output to a whitelist of keys {0, ..., 87}.
    * @param lastOutput Override for last note output. If unspecified, uses 
    * output from the last call to this method.
    * @param deltaTime Override for elapsed time since last button press. If
```
notwaldorf: could we split this into functions with a more manageable number of parameters? 7 is a bit much, and since they're not named you can start writing code like this which is not great: `next(1, null, null, [1,2,3,4...80], null 4)`

PR #198, opened 2018-11-21
Add arbitrary style transfer model

reiinakano:
Here's a quick initial implementation. Please let me know what you think. I copied most of the config files over from sketch. It seems to build fine. 

Some notes:
Please let me know if my typescript is correct, I haven't done much TS.
I guess I should default to the GCP checkpoints?
Please let me know what you think is the best way to test this code.

Oh and btw you may notice border artifacts in the stylized images. This is because I had to switch out the reflection padding layers with zero padding layers. As soon as https://github.com/tensorflow/tfjs/issues/900 is fixed, I can switch it out.

I will go to bed now as it's 3AM here :) I'll probably get back to you guys by tomorrow.

@adarob @vdumoulin


arbitrary-image-stylization/src/style/model.ts
```diff
@@ -0,0 +1,130 @@
...
+    this.dispose();
+
+    [this.styleNet, this.transformNet] = await Promise.all([
+      tf.loadFrozenModel(
+        this.styleCheckpointURL + '/tensorflowjs_model.pb',
+        this.styleCheckpointURL + '/weights_manifest.json'
+      ),
+      tf.loadFrozenModel(
+        this.transformCheckpointURL + '/tensorflowjs_model.pb',
+        this.styleCheckpointURL + '/weights_manifest.json'
```
vdumoulin: Typo?
```suggestion
        this.transformCheckpointURL + '/weights_manifest.json'
```

arbitrary-image-stylization/src/style/model.ts
```diff
@@ -0,0 +1,130 @@
...
+import * as tf from '@tensorflow/tfjs';
+
+/**
+ * Main ArbitraryStyleTransferNetwork class
+ */
+export class ArbitraryStyleTransferNetwork {
+  private styleCheckpointURL: string;
+  private transformCheckpointURL: string;
+
+  private initialized: boolean;
```
vdumoulin: Should this be initialized?
```suggestion
  private initialized: boolean = false;
```

vdumoulin:
Additional comments:

> I guess I should default to the GCP checkpoints?

Yes, I think that would be a sane default.

> Please let me know what you think is the best way to test this code.

Having tried it out, I feel fairly confident that it works as expected (modulo the suggested changes). @adarob are there additional tests or verifications you think would be relevant?

> Oh and btw you may notice border artifacts in the stylized images. This is because I had to switch out the reflection padding layers with zero padding layers. As soon as tensorflow/tfjs#900 is fixed, I can switch it out.

I suspected that was the case. IMO that's absolutely not a blocker, but it's good that you pointed it out. Let us know when you have an updated version of the model, and we'll handle swapping the files on GCS.

adarob:
We should actually call this new library @magenta/image. Please update the package.json, etc., and  change the path to be image/arbitrary-stylization.

arbitrary-image-stylization/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "description": "Arbitrary image stylization in the browser.",
+    "main": "es5/index.js",
+    "types": "es5/index.d.ts",
+    "jsdelivr": "dist/magentastyle.js",
+    "unpkg": "dist/magentastyle.js",
+    "dependencies": {
+        "@tensorflow/tfjs": "^0.13.3"
+    },
+    "devDependencies": {
+        "@types/clone": "^0.1.30",
```
adarob: Please remove packages you aren't using. I think this would include tape, clone, and fs.

adarob:
@reiinakano we will pull this in as-is and then refactor as necessary. Thanks for the contribution!

PR #217, opened 2018-12-06
Add documentation and demo for `image` library

vdumoulin:


image/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Contents
+
+- [Example Applications](#example-applications)
+- [Supported Models](#supported-models)
+- [Getting Started](#getting-started)
+
+## Example Applications
+
+Here are a few applications built with `@magenta/image`:
```
notwaldorf: you can remove this section until you have a demo to link to, wdyt?

image/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+You can also try our [hosted demos](https://tensorflow.github.io/magenta-js/image/demos) for each model and have a look at the [demo code](./demos).
+
+## Supported Models
+
+### Fast Arbitrary Image Stylization
+
+Implements Ghiasi et al.'s fast arbitrary style transfer model ([paper](https://arxiv.org/abs/1705.06830), [code](https://github.com/tensorflow/magenta/tree/master/magenta/models/arbitrary_image_stylization)). Wraps around Reiichiro Nakano's [TensorFlow.js port](https://github.com/reiinakano/arbitrary-image-stylization-tfjs) of the model checkpoint.
+
+**Demo Application:** Coming soon!
```
notwaldorf: same

image/README.md
```diff
@@ -0,0 +1,119 @@
...
+    <img id="style" src="style.jpg"/>
+    <canvas id="stylized"></canvas>
+      <script>
+        const model = new mi.ArbitraryStyleTransferNetwork();
+        const contentImg = document.getElementById('content');
+        const styleImg = document.getElementById('style');
+        const stylizedCanvas = document.getElementById('stylized');
+
+        async function stylize() {
+          const bottleneck = await mi.tf.tidy(() => {
```
notwaldorf: why is this called bottleneck? is there a more descriptive word, maybe? like `styleParameters`? (i know nothing about this model)

image/demos/style.css
```diff
@@ -0,0 +1,71 @@
...
+button[disabled] {
+  opacity: 0.5;
+}
+
+.image {
+  margin-right: 10px;
+  outline: 3px solid #222;
+}
+
+.player-container {
```
notwaldorf: heh, you probably don't need player-container and visualizer-container. they are very music specific 😅

notwaldorf:
looks super good! mostly nits, tbh

image/README.md
vdumoulin: Done.

image/README.md
vdumoulin: Done.

image/README.md
vdumoulin: That's the language used in the paper (this corresponds to the output of a "bottleneck" layer just before the linear projection onto the instance normalization parameter values), but I don't feel very strongly about keeping that name and agree that a more descriptive name would be better.

I changed the name to `styleRepresentation`.

image/demos/style.css
vdumoulin: Oops, my bad! Done.

adarob:
LGTM, but I think we need to actually publish the npm first?  Vincent, do you have an npmjs account I can add to the magenta org?

vdumoulin:
@adarob I do as of 5 minutes ago. Same username: `vdumoulin`.

image/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+### via Script Tag
+
+Add the following code to an HTML file, and place a content (`content.jpg`) and style (`style.jpg`) image in the same directory:
+
+```html
+<html>
+  <head>
+    <!-- Load @magenta/image -->
+    <script src="https://cdn.jsdelivr.net/npm/@magenta/image@^1.0.0"></script>
```
adarob: It looks like you published 0.1.0, not 1.0.0. Maybe just change this number?

adarob: Also, verify this actually works now that we've published the package.

PR #220, opened 2018-12-06
add mm.Visualizer autoscrolling

notwaldorf:
This basically gets copy pasted for every usage of the Visualizer and that's silly.


music/src/core/visualizer.ts
```diff
@@ -111,11 +113,14 @@ export class Visualizer {
    * active
    * @param activeNote (Optional) If specified, this `Note` will be painted
    * in the active color.
+   * @param scrollIntoView (Optional) If specified and the note being painted is
+   * offscreen, the parent container will be scrolled so that the note is
+   * in view
```
adarob:
```suggestion
   * in view.
```

notwaldorf:
thanks!!

PR #226, opened 2018-12-14
SketchRNN cleanup

notwaldorf:
This is a bit of a big'un. The major change is that I rewrote the demos to be more clear, and use more standard p5.js code (the previous `p.draw` functions were a bit terrifying since they did everything from handling user drawing and model drawing), and use actual DOM elements instead of drawing them on a canvas. I've comapred the demos to the best of my abilities to the existing ones, they look ok?

I've also cleaned up the readme a bit, and added a `publish-demos` to the `package.json`, so that we can host the demos like we do for `magenta/music`


notwaldorf:
(im gonna wait to see if @hardmaru has any comments before merging this)

PR #238, opened 2018-12-31
Fix blurry notes when using the Visualizer

Gerrit0:
The calculated position of notes aren't always integers, which results in a blurry visualization. 

See https://stackoverflow.com/q/53918770/7186598 for more details.

### DIFF
--- a/music/src/core/visualizer.ts
+++ b/music/src/core/visualizer.ts
@@ -127,6 +127,7 @@ export class Visualizer {
     // active notes instead.
     this.ctx.clearRect(0, 0, this.ctx.canvas.width, this.ctx.canvas.height);
     let activeNotePosition;
+    const noteRenderHeight = Math.round(this.config.noteHeight);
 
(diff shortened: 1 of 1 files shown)


music/src/core/visualizer.ts
```diff
@@ -152,7 +152,9 @@ export class Visualizer {
       this.ctx.fillStyle =
           `rgba(${isActive ? this.config.activeNoteRGB : this.config.noteRGB},
           ${opacity})`;
-      this.ctx.fillRect(x, y, w, this.config.noteHeight);
+      // Round values to the nearest integer to avoid partially filled pixels.
+      this.ctx.fillRect(Math.round(x), Math.round(y), Math.round(w),
+        Math.round(this.config.noteHeight));
```
notwaldorf: Could we maybe round the noteHeight once, outside of the loop? It’ll be the same through and there’s no point rounding it over and over again.

notwaldorf:
Thanks so much!

PR #240, opened 2019-01-03
Publish magenta/sketch demos

notwaldorf:
(and bump to 0.1.4)


notwaldorf:
@adarob The links are already updated from before: https://github.com/tensorflow/magenta-js/blob/master/sketch/README.md#1-simplehtml--simplejs

PR #241, opened 2019-01-04
Webpack dev-server

cannoneyed:
Switches the default build/serve story to use a webpack dev server, for automatic live reload and faster development cycles. Additionally, allows the user to build specific demos via a CLI argument.

### DIFF
--- a/music/package.json
+++ b/music/package.json
@@ -31,6 +31,7 @@
     "fs": "^0.0.1-security",
     "http-server": "^0.11.1",
     "in-publish": "^2.0.0",
+    "minimist": "^1.2.0",
     "tape": "^4.9.0",
     "ts-loader": "^5.3.0",
     "ts-node": "^5.0.1",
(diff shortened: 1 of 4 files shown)


notwaldorf:
I'm not super familiar with webpack in general, as you know. Is the advantage to this that you don't need to run the build script every time you make a change, and it watches the files for you instead?

cannoneyed:
Hey!

Yeah, the advantage is that you get watch / reload on change for the demo development. This is super useful for the work I've been doing on them! Also, being able to specify building specific demos drastically increases the build speed.

notwaldorf:
> being able to specify building specific demos drastically increases the build speed.
this used to work when we had the build script already. did this break when we switched to webpack?

music/package.json
```diff
@@ -50,7 +52,7 @@
     "test": "ts-node node_modules/tape/bin/tape src/**/*_test.ts",
     "build-demos": "webpack --config ./webpack/demo.config.ts",
     "publish-demos": "yarn build-demos && mkdir -p ../docs/music/demos && cp demos/*.{js,html,mid,css} ../docs/music/demos",
-    "run-demos": "yarn build-demos && yarn serve-demos",
+    "run-demos": "webpack-dev-server --config ./webpack/demo.config.ts",
     "serve-demos": "http-server demos/",
```
notwaldorf: is 'serve-demos' still needed?

cannoneyed: probably not? I wouldn't use it, but I wasn't sure if others would rather serve the built bundles rather than rely on webpack.

adarob:
> This worked super neat when I ran it! @adarob wanna give it a try too?

I will try it once it's checked in, thanks!

notwaldorf:
Coolio, I'll merge and not do a release then, so you have a chance to double check it

PR #247, opened 2019-01-10
Regenerate docs

notwaldorf:
There was a typing error (that somehow the linter didn't catch but `typedoc` was really upset about), which made `typedoc` just exit huffing and puffing and docs stopped getting generated.


docs/music/classes/_core_chords_.chordencoder.html
```diff
@@ -233,6 +233,9 @@ <h4 class="tsd-returns-title">Returns <span class="tsd-signature-type">Tensor</s
 					<li class=" tsd-kind-external-module">
 						<a href="../modules/_core_visualizer_.html">"core/visualizer"</a>
 					</li>
+					<li class=" tsd-kind-external-module">
```
notwaldorf: @adarob any idea why this has started appearing?

notwaldorf: oh i got it, it's from the webpack change. lemme blacklist it too.

PR #251, opened 2019-01-17
Changes recorder note timings to use midi event times when available …

kmalta:
…and falls back to performance.now timings when not.

### DIFF
--- a/music/src/core/recorder.ts
+++ b/music/src/core/recorder.ts
@@ -293,9 +293,11 @@ export class Recorder {
     // event.timeStamp doesn't seem to work reliably across all
     // apps and controllers (sometimes it isn't set, sometimes it doesn't
     // change between notes). Use the performance now timing, unless it exists.
-    let timeStampOffset = performance.now();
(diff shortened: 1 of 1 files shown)


music/src/core/recorder.ts
```diff
@@ -292,8 +292,12 @@ export class Recorder {
 
     // event.timeStamp doesn't seem to work reliably across all
     // apps and controllers (sometimes it isn't set, sometimes it doesn't
-    // change between notes). Use the actual message time for now.
-    const timeStamp: number = Date.now();
+    // change between notes). Use the performance now timing, unless it exists.
+    let timeStampOffset = performance.now();
+    if (event.timeStamp !== undefined && event.timeStamp !== 0) {
```
notwaldorf: Actually, I'd prefer this to not be set if we do have the message time stamp. I.E.:
```
let timeStampOffset;
if (event.timeStamp !== undefined && event.timeStamp !== 0) {
...
} else {
timeStampOffset = performance.now();
}
```

PR #254, opened 2019-01-23
Make recorder timing more accurate

notwaldorf:
Good news, I finally understood what @tambien meant in [this comment](https://github.com/tensorflow/magenta-js/pull/181#discussion_r224963796), so here's that change.


notwaldorf:
cc @kmalta

PR #255, opened 2019-01-23
Updates the callbackobject for the Recorder object to include noteOn and noteOff events.

kmalta:


music/src/core/recorder.ts
```diff
@@ -49,6 +49,24 @@ export abstract class BaseRecorderCallback {
...
   abstract run(seq: NoteSequence): void;
+  /**
+   * Will be called for each time a note on event is observed.
+   *
+   * @param pitch The pitch of the midi event received.
+   * @param velocity The velocity of the midi event received.
+   * @param device The device the midi event was received from.
+   */
+  // tslint:disable-next-line:no-any
+  abstract noteOn(pitch: number, velocity: number, device: any): void;
```
notwaldorf: I think you can use the proper type here, `WebMidi.MIDIInput` maybe?

music/src/core/recorder.ts
```diff
@@ -315,14 +333,21 @@ export class Recorder {
     const cmd = event.data[0] >> 4;
     const pitch = event.data[1];
     const velocity = (event.data.length > 2) ? event.data[2] : 1;
+    const device = event.srcElement;
 
     // Some MIDI controllers don't send a separate NOTE_OFF command.
     if (cmd === NOTE_OFF || (cmd === NOTE_ON && velocity === 0)) {
+      if (this.callbackObject) {
```
notwaldorf: `callbackObject` might be defined, but `noteOn`/`noteOff` might not be (outside of typescript nobody will yell at you), so I would check if those exist as well.

notwaldorf:
@kmalta Is this ready for review? The tests are still failing...

PR #259, opened 2019-02-01
Add a split() sequence helper

notwaldorf:
So that you can split a sequence into chunks (useful if whatever models you're using are a small number of bars, and you want to encode a very long melody in them)


notwaldorf:
Yikes, actually this doesn't work in the code i'm using it in so uhhh maybe hold off on review.

music/src/core/sequences.ts
```diff
@@ -424,3 +424,95 @@ export function mergeInstruments(ns: INoteSequence) {
...
+ *
+ * @param ns The `NoteSequence` to split.
+ * @param chunkSize The number of steps to split the note sequence in.
+ * For example, if you want to split the sequence into 2 bar chunks, then
+ * if the sequence has 4 steps/quarter, that will be 32 steps for each 2 bars.
+ *
+ * @returns An array of `NoteSequences` each of which are at most `chunkSize`
+ * steps.
+ */
+export function split(seq: INoteSequence, chunkSize = 32): NoteSequence[] {
```
adarob: i think chunkSize should not have a default

music/src/core/sequences.ts
```diff
@@ -424,3 +424,95 @@ export function mergeInstruments(ns: INoteSequence) {
...
+      continue;
+    }
+    // If this note fits in the chunk, add it to the current sequence.
+    if (note.quantizedEndStep <= chunkSize) {
+      currentNotes.push(note);
+    } else {
+      // If this note spills over, truncate it and add it to this sequence.
+      if (note.quantizedStartStep < chunkSize) {
+        currentNotes.push(new NoteSequence.Note({
+          pitch: note.pitch,
```
adarob: I worry that you may only copy a subset of the fields, for example if we add another later. Can you just copy the whole note and then change the steps?

music/src/core/sequences.ts
```diff
@@ -424,3 +424,95 @@ export function mergeInstruments(ns: INoteSequence) {
...
+export function split(seq: INoteSequence, chunkSize = 32): NoteSequence[] {
+  assertIsQuantizedSequence(seq);
+
+  // Make a clone so that we don't destroy the input.
+  const ns = clone(seq);
+
+  // Sort notes first.
+  const notesBystartStep =
+      ns.notes.sort((a, b) => a.quantizedStartStep - b.quantizedStartStep);
+
```
adarob: We should probably explicitly clear all of the absolute times since they will no longer be valid.

music/src/core/sequences.ts
```diff
@@ -424,3 +424,95 @@ export function mergeInstruments(ns: INoteSequence) {
...
+        note.quantizedStartStep = originalStartStep;
+        note.quantizedEndStep = originalEndStep;
+      }
+
+      // Do we need to look at this note again?
+      if (note.quantizedEndStep > chunkSize ||
+          note.quantizedStartStep > chunkSize) {
+        i = i - 1;
+      }
+      // Save this bar if it isn't empty.
```
adarob: s/bar/chunk/g

music/src/core/sequences.ts
notwaldorf: did below (when creating the new note)

music/src/core/sequences.ts
```diff
@@ -424,3 +424,95 @@ export function mergeInstruments(ns: INoteSequence) {
 
   return result;
 }
+
+/**
+ * Splits an unquantized `NoteSequence` into chunks.
+ *
```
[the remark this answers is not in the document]
notwaldorf: Is this better?

PR #260, opened 2019-02-01
Make a mm.visualizer that uses svg instead of canvas

notwaldorf:
Because of performance reasons, if you have other memory intensive things going on the page (tfjs, i'm looking at you), some very long midi files (especially if they're quantized) lead to super wide canvases (like 15k pixels), and the regular `mm.Visualizer` literally crashes (and displays nothing).

This is a super quick visualizer that has the same API, only uses an SVG instead of canvas. I've also tried to have a good-ish base class so that new visualizers can be implemented from it.

Other changes:
-  renamed `mm.Visualizer` to `mm.PianoRollCanvasVisualizer` to be more clear (since now there's >1 visualizers), and added a deprecation warning to it.
- updated all the demos to use the new `mm.PianoRollSVGVisualizer` since it's moar better.


rogerpasky:
@notwaldorf please notice those PR changes does not include the former ones you did in **mm.Visualizer: use a better default for the pixels per unquantized step** #258

Given there are some relevant code restructuration due to new base classes, I guess merge won't be trivial to be done by the automatic merger and could be overwritten by the old pixel defaults.

notwaldorf:
Yup, no worries, I didn’t get to the merge yet

music/src/core/visualizer.ts
```diff
@@ -121,118 +242,200 @@ export class Visualizer {
    * active
    * @param activeNote (Optional) If specified, this `Note` will be painted
    * in the active color.
-   * @param scrollIntoView (Optional) If specified and the note being painted is
+   * @param scrollIntoView (Optional) If specified and the note being painted
+   *     is
```
adarob: check this formatting

PR #269, opened 2019-02-20
Update tfjs to latest version

notwaldorf:
Updates `tfjs` to the latest version. This was a bit of a pain and involved:
- fixing a bunch of TypeScript errors.
- ~*adding a p dubious `tf.disableDeprecationWarnings();` to the tests, to supress a warning that says "Tensor.get() is deprecated. Use Tensor.array() and native array indexing instead.". I'd love to fix it, but I don't understand who is generating that warning. I've added it to a test file that has two `.get()` calls, but if I comment out those tests then the errors still get generated. *~

I've also changed the demos to import `tfjs` instead of `tfjs-core`, for consistency. I don't know if it makes a difference, and I can revert this change if there was a good reason to have it there.

I've also cleaned up the `package.json` a bit while I was there:
- we were depending on a `fs` package that doesn't exist
- we were depending on an old `tfjs` package that didn't seem to be used

### DIFF
--- a/music/src/core/data_test.ts
+++ b/music/src/core/data_test.ts
@@ -16,6 +16,9 @@
  */
 
 import * as tf from '@tensorflow/tfjs';
+
+// To remove the warning that tf.get() is deprecated.
+tf.disableDeprecationWarnings();
 import * as test from 'tape';
 
 import {NoteSequence} from '../protobuf/index';
--- a/music/src/core/data_test.ts
+++ b/music/src/core/data_test.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import * as tf from '@tensorflow/tfjs-core';
+import * as tf from '@tensorflow/tfjs';
 import * as test from 'tape';
 
 import {NoteSequence} from '../protobuf/index';
--- a/music/src/core/data_test.ts
+++ b/music/src/core/data_test.ts
@@ -16,11 +16,6 @@
  */
 
 import * as tf from '@tensorflow/tfjs';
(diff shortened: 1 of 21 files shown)


music/src/core/data_test.ts
```diff
@@ -15,7 +15,10 @@
  * limitations under the License.
  */
 
-import * as tf from '@tensorflow/tfjs-core';
+import * as tf from '@tensorflow/tfjs';
+
+// To remove the warning that tf.get() is deprecated.
```
adarob: why not just stop using tf.get?

notwaldorf: That's what I was trying to explain in the PR description. I can only find 2 instances of `.get()`, in this file. Even if I remove those tests, I still get all the errors (read: hundreds), and I can't find where the other instances are. I searched in files for any `.get(`, and we only use it to get from a `Map`....

notwaldorf: Ok so I tried to replace the two `.get()` calls in this function but `arraySync` seems to be busted for scalars, so I filed https://github.com/tensorflow/tfjs/issues/1271.

notwaldorf: Oh apparently 0.15.3 fixes this so uhhh hold.

notwaldorf: fixed!

notwaldorf:
@adarob fixed the deprecation warnings and removed the TODO, ptal? 🙏

notwaldorf:
thanks!

Issue #246, opened 2019-01-09
Add ScoreVisualizer to display NoteSequences as musical scores in staffs


rogerpasky:
Current `Visualizer` class is cool (great job @notwaldorf ), and it could evolve to display sequences not just as pianoroll, but **music staff** as well. A derived `ScoreVisualizer` is proposed.

It must fit current pianoroll functionality in a polymorphic way introducing minimal changes:
- **`constructor(sequence, canvas, config)`** which:
  - Constructs all the internals according to the given config.
  - Ends up doing the initial redraw.
- **`redraw(activeNote, scrollIntoView)`** which:
  - Uses the graphical placeholder (currently `canvas`) to do the drawing (if needed) with the non-highlighted style.
  - Uses the `activeNote`, if given, to highlight it in the drawing, dimming the style colour according with the velocity note (if any).
  - Scrolls the graphical placeholder to locate the `activeNote` properly if requested in `scrollIntoView`.

Minimum Viable Product (MVP) must add new music staff functionality through the same `Visualizer` interface:
- [x] Prepare a class hierarchy to allow SVG visualizers.
- [x] Provided values `config.minPitch` and `config.maxPitch` should be used to select the musical clef (at least treble clef and bass clef). Undetermined cases will default to treble clef.
- [x] Different values of time signature will be extracted from given `sequence`, if any. Otherwise it will be assumed 4/4 for all staff.
- [x] It will be displayed the clef signature at the beginning of the staff, and it won't change through the score.
- [x] No key signature will be displayed at the beginning of the staff. It will be assumed C key. 
- [x] Initial time signature will be displayed at the beginning of the staff.
- [x] Bar separators will be displayed according with initial time signature.
- [x] There will only be a single five-line staff per `NoteSequence`. Multiple staff will be handled instantiating several vertically aligned `ScoreVisualizer` instances with different `config.minPitch` and `config.maxPitch` values.
- [x] Horizontal note disposition will be coherent with current pianoroll `Visualizer` in order to allow aligning heterogeneous visualizers vertically. It's out of scope in this class to visualize compact staff which would be handled in future derived visualizers.
- [x] Musical clef, time signature and key signature displayed at the beginning of the staff will be handled to allow vertical alignment with other visualizers as mentioned. Yet to be detailed.
- [x] Notes and rests will be drawn from whole note (semibreve) to sixty-fourth note (hemidemisemiquaver) including dotted-note versions (single dotted). 
[...]

PR #280, opened 2019-03-11
StaffSVGVisualizer Minimum Viable Product #246

rogerpasky:
All proposed features and some extras have been developed. Note ties should be included in order to accept any multi track score, so simultaneous notes with different starting or ending would be drawn without breaking anything.

Scrolling in the base class have been adapted to do a smoother movement note by note because some problems were found due to padding size was not had in consideration.

All SVGs were drawn by myself and donated to the project, so no copyright problems involved.


notwaldorf:
Hiya! I was on vacation and dropped the ball on this. Will try to review this week!

rogerpasky:
Hi @notwaldorf !!! No worries, we all work exclusively to enjoy vacation time ;-)

I'm pretty sure you'll love the result:
image

Notice there are two visualization modes available: "proportional", which distributes note positions according to the note length (as shown in the previous image) and "compact" which draws notes as a printed score using a standard padding among notes.

I've been advancing in the development, so **I'm about to commit multi-clef and multi-time signature** features which would allow the visualizer to draw almost any score (letting apart just triplets and beams). I'll include an intense refactor to ease maintenance. **This will be done in a couple of days, so there's no need to rush**.

BTW, just to let you know, I've noticed some failures in mm.PianoRollSVGVisualizer:
- SVG width is wider than score, adding a gap at the right.
- Due to JavasCript floating point number inaccuracies, in polyphonic scores I've noticed there can appear notes which are kept as "Active Note" even though they are not.
image

First one is not important at all, but I can fix the second one if you want. Just let me know.

notwaldorf:
Oh yeah, I've definitely noticed the weirdness with the SVG visualizer. If you have a fix and would like to send it as a separate PR I would more than appreciate it 🙏

rogerpasky:
You can count with it! I'll create a new issue for the record and following I'll do the PR.

I'll probably delay my own PR till Sunday to make room for extra-checkings.

rogerpasky:
Found and solved @notwaldorf !

- On top of JS float inaccuracy, you were removing just one single active note (added a while loop).
- The extra width space was caused by the accumulation of extra spacing inter-notes, which was changed for a subtraction in the drawing but not in the SVG / canvas general width calculation.
- I've found and solved another one linked to the correction you did to make same widths on quantized and non-quantized sequences. Your fix worked just for players using default values, but not with the ones who were giving a `config.pixelsPerTimeStep`value to the constructor.
- The last one is related to the scroll. Horizontal paginating had a problem because of the DIV padding size which was not taken into account and ended up out of focus. I've changed it to maintain the active note in the middle of the DIV, and added an extra CSS value (` scroll-behavior: smooth;`) to make it more "pro" (yep, I know this one is not standard yet, but works perfectly in Chrome and Firefox, and doesn't break anything on Safari).

If you agree with the scope, I'll create directly a new PR (apart from the one I'm working on) with the related description in order to push it from your side.

PS: Congrats for today's Bach Doodle, I love it!!!

notwaldorf:
Those fixes sounds wonderful and very correct. I did a sloppy job porting/testing that SVG visualizer 🤦‍♀️! Send them over!

(thanks for the doodle! we're super excited about it!!)

rogerpasky:
Help @notwaldorf I need some advice to do the separate PR.

Github notices me I already have done a fork to do a PR (this one) and doesn't let me do another one. As far as I know (I'm pretty new contributing other's repos and doing PRs) I can do:

- Save a copy of my code, undo / cancel the PR and start it over with the fixes of your code and, once they get approved, restart again with the saved StaffSVGVisualizer files in a new fork/PR.
- Continue with my deep testing and push the changes to this very PR as soon as possible (I was meant to finish on Sunday, but I could accelerate it for tomorrow).

It there's a third way I'll be more than happy to follow it.

notwaldorf:
Oh, so on the same fork you can have multiple branches (otherwise I wouldn't be able to get any work done 😅 . On the command line, you can run `git checkout -b fix-svg-viz` or whatever, and then just send that branch as a PR. 

Here's the github help page on branches: https://help.github.com/en/articles/about-branches. Let me know if this doesn't make sense!

music/demos/visualizer.ts
```diff
@@ -91,3 +101,243 @@ function startOrStop() {
     playBtn.textContent = 'Stop';
   }
 }
+
+
+// Section 2 ******
+
+const ADVANCED_MIDI_URL = './anna-magdalena-bach.mid';
```
notwaldorf: I wouldn't include a midi with this PR. We already have the option of loading you own midi, and i think that's enough

music/demos/visualizer.ts
```diff
@@ -70,8 +75,13 @@ function initPlayerAndVisualizer(seq: mm.INoteSequence) {
   playBtn.disabled = false;
   playBtn.textContent = 'Loading';
 
-  canvasVisualizer = new mm.PianoRollCanvasVisualizer(seq, canvas);
-  svgVisualizer = new mm.PianoRollSVGVisualizer(seq, svg);
+  let config = {
```
notwaldorf: is this needed? does everything look ok if it's not set?

music/demos/visualizer.html
```diff
@@ -50,21 +51,87 @@ <h1>Visualizers</h1>
...
+    <div class="visualizer-container" id="container3">
+      <canvas id="piano-roll-canvas-viewer"></canvas>
     </div>
   </section>
 
-  <h2>mm.PianoRollCanvasVisualizer</h2>
   <section>
-    <div class="visualizer-container" id="container">
-      <canvas id="canvas"></canvas>
+    <h2>Advanced usage</h2>
```
notwaldorf: This entire paragraph feels like it belongs in the class docs for the SVGVisualizer, not the demo. For the demo I would maybe keep the multiple key signature example, since that seems different and maybe not straightforward?

rogerpasky:
Please remember this PR is prior to the other one and doesn't include recent changes.

rogerpasky:
The "keep the active note in the middle of the screen" is optional now. By default it follows your pagination and it has a third mode which is halfway, doing pagination by bar, as you can see in the Bach score (actually it wasn't composed by good old Johan Sebastian).

rogerpasky:
The notes directly to the right of a bar line are very squished in non-compact visualization mode because it was the only place to avoid the bar colliding with previous notes. Actually the bar is drawn before the note, not the note after the bar. Think about the notes in non-compact mode as your rectangles, placed proportionally at the same where the leftmost side of the rectangle would be, so , if you would have drew bars in the piano roll visualizer, they would overlap, or even worse, stay before previous notes. I can ensure this is even more confusing than having it where they are (I've tested tens of variants). Even I managed to get the bar exactly in between the note head and the preceding accidental (sharp, flat...) to avoid overlappings.

Nevertheless, If there's a need of paddings, there it is the compact visualization mode as another option.

My experience is the former is useful and plenty of sense in multi staff scores and the latter is more suitable for single staff scores. Plenty of choices!!! :-D

music/demos/visualizer.ts
rogerpasky: I think it's important to have a medium complexity score in order to test and show the visualizer capabilities in the demo. If there's a problem about serving the MIDI, I would convert it to ProtoBuf in a separate .ts file and include it like Twinkle tune. It would help to handle manually some leaks in the MIDI conversor (like not handling the key signature of the MIDI file).

music/demos/visualizer.ts
rogerpasky: Currently under revision. Let's review this code when it is done in next PR.

rogerpasky:
Finally I got it done @notwaldorf !!! It has been a very deep refactoring and deep testing (weeks) to verify the infinite number of use cases (including combinations). I tried to explain it all in the demo page as sort of a tutorial. I hope you like the result as much as I do.

rogerpasky:
Regarding to the "blinking thing", which is not a bug, but intentional, I've modified it to be more visually appealing: animation is longer, smoother with easy in/out, and spends more time in the "score side" than the "signatures side".

I've dedicated a lot of time thinking about how to represent the signatures in proportional mode till I found this solution as the best possible discarding all others: putting a scaled version of signatures over the staff was not intelligible, making it appear on-hover was not practical for reading during play, no representation was no acceptable given it provides a lot of semantic to following symbols, etc. If you think there's a better way to represent it, please, let me know.

Please notice the example you captured (narrow scale of 16th notes) is the greatest counter-example of its usefulness. Please have a look at the multi-signature example to realize all its potential (click several times to [+] button to add different signatures and do scroll to see it in action).

rogerpasky:
I've done some extra-heavy testing visualizing Chopin "Heroic" Polonaise (very complex score, 54KB MIDI file) and complete Bach Goldberg Variations (extra large 250KB MIDI file) and found an extra use case, which is the handling of notes shorter than the lowest resolution (1/64th). Fixed it and it works perfectly, even with misconfigured MIDI files handling zero-length notes.

Canvas visualizer crashed on long score as expected, but both SVG visualizers did a perfect job!!!

music/demos/visualizer.ts
notwaldorf: I loaded Aladdin with these settings and it looks pretty bad, so I don't think we want to keep them like this:
Screen Shot 2019-04-12 at 11 01 31 AM

music/demos/visualizer.html
```diff
@@ -50,21 +51,148 @@ <h1>Visualizers</h1>
...
     </div>
   </section>
 
-  <h2>mm.PianoRollCanvasVisualizer</h2>
   <section>
-    <div class="visualizer-container" id="container">
-      <canvas id="canvas"></canvas>
+    <h2>mm.StaffSVGVisualizer advanced usage</h2>
+    <p>
+      mm.StaffSVGVisualizer extends the basic configuration capabilities to add:
```
notwaldorf: Again, as I mentioned before, I think these belong as class comments, not as comments in this demo.

music/demos/visualizer.html
```diff
@@ -50,21 +51,148 @@ <h1>Visualizers</h1>
...
+        </li>
+        <li>
+          Follow scoreplaying with two added auto-scroll modes, besides default pagination  
+          (<b>ScrollType.PAGE</b> or void), like centering active note on the middle of the 
+          score in a note by note basis (<b>ScrollType.NOTE</b>) or centering it in bar sized 
+          packed chunks (<b>ScrollType.BAR</b>) every time activeNote is the first one of a
+          bar.
+        </li>
+        <li>
+          Multiple <b>key signature</b> and <b>time signature</b> are supported (but notice
```
notwaldorf: I would add this as a comment and not include it in the demo, then. Most of the visualizations Magenta users do are of MIDI files, and if this is not automatic then it's not really useful to call it out as a demo. You can call it out as a feature in the class docs if you want, but I think it's very confusing to have it here.

music/demos/visualizer.html
```diff
@@ -50,21 +51,148 @@ <h1>Visualizers</h1>
...
+
+    <div class="visualizer-container" id="container4">
+      <div id="right-hand"></div>
+      <div id="left-hand"></div>
     </div>
-  </section>
+    <br>
+
+    <h3>Multiple Key and Time signatures</h3>
+    Next example adds more notes to the score (+) changing through all keys and the time signatures
```
notwaldorf: As mentioned before, I don't think this demo belongs here

music/demos/visualizer.html
```diff
@@ -50,21 +51,148 @@ <h1>Visualizers</h1>
...
+      <span id="tempoValue3"></span>
+    </span>
+    <br>
+    <br>
+    <div class="visualizer-container" id="container5">
+      <div id="multiple-signatures"></div>
+    </div>
+
+    <h3>Musical symbol catalog and managed use cases</h3>
+    Next score shows all musical symbols (but F clef). In a G clef score, C key and 4/4 time signature,
```
notwaldorf: I would include this if possible in the main demo, or not include it at all (I think people expect that notes on this staff visualizer should work as expected, so we don't need to demonstrate this necessarily)

music/src/core/visualizer.ts
```diff
@@ -441,3 +441,1266 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+ * the right accidentals. It can be overwritten with 
+ * `NoteSequence.keySignatures` value at time or step 0. If not assigned it 
+ * will be asumed C key.
+ * @param instruments The subset of the `NoteSequence` instrument track numbers 
+ * which should be merged and displayed. If not assigned or equal to [] it will 
+ * be used all instruments altogether.
+ * @param scrollType Sets scrolling to follow scoreplaying in different ways 
+ * according to `ScrollType` enum values.
+ */
+export interface AdvancedVisualizerConfig extends VisualizerConfig {
```
notwaldorf: You can just call this StaffVisualizerConfig, since it's not used by anything else

music/demos/visualizer.ts
rogerpasky: It has been revised and it is Ok. The code I guess you miss (the removed lines 73 and 74, doing the object construction) are a few lines below (82, 83 and 84). Actually they have not been translated, but merger algorithm did not match them properly. They moved below precisely due to the addition of a config.pixelsPerTimeStep setup to 60 pixels/second, because default value of 30px is too narrow to make room for note head width.

Regarding to the ugly aspect, I do agree such a fast score with lots of 1/64th, several tracks and just 60px per second width resolution looks horrible in proportional mode. It will improve if I force the visualizer to draw in compact mode on this first part of the demo (all visualizers following same player).  I can make this configuration change to 0px/second, which is the value to make it compact visualization, and apply it only to the mm.ScoreSVGVisualizer instance (the rest will keep default values as always).

I did it so because I wanted to show how well they matched (horizontally) checking the compatibility among 3 visualizers, but I can move it to a separate demo as you suggested.

music/demos/visualizer.html
rogerpasky: There is not an "as expected" musical score standard. There are several interpretations of the different ways of doing (same as you can get form "standards" from different musical SW like Finale, Cubase, Logic...), and not just in the visual point of view, but in the musical rules interpretation, which some of them are subject to composer intention among many choices. This is a way to represent "what you will get visually", which cannot be shown in class comments.

I agree It could get out of the general Visualizers demo and go to a separate demo page.

music/src/core/visualizer.ts
rogerpasky: Sure! It was the original name and I was planning to change it exactly as you mention, but I forgot it. Thanks.

rogerpasky:
Ok I do agree.

> The blinking should not be the default mode. Compact mode should be the default, and if users want to change from that, they should. Out of the box, the visualizer should be in the least surprising mode, and the blinking is very confusing.

I'm changing it to make `config.pixelsPerTimeStep` 0px as default value, which means "class will handle the note position and padding automatically". This will show no blinking unless you intentionally set a greater than zero value. Having both options is great, and I don't mind which is the default one.

> I don't think the demo is in the right shape right now. The demo should just be an example of what this does, and not act as the docs for the class. I think a lot of what's in the demo should be class comments.

There are plenty of internal class comments, but I given this is a dynamical front-end functionality, there's a need to see it working properly. More in next quote answer.

> In particular, I'm not sure the demo should show things that aren't possible with MIDI. Most of the usage we get out of the visualizers is from loading a midi into them -- a lot of the features you're showing are for constructing the staff visualization by hand. Maybe this should be a separate demo page, if you think that it's a key feature.

I respectfully disagree with that. I've done this visualizer to see the result of Magenta models output. I'm not willing to construct it by my hand, but by computer's hand. That's Magenta all about, isn't it? The things I want to do by hand I do with a decent musical editor, like Finale or Logic :-) .

But MIDI is not the one who limits the features, but the Magenta `ProtoBuf`implementation, which has some failures I found digging into the code I'll report as issues. They are related to the duality time / quantization which is not consistently spread all over ProtoBuf fields, mainly in signature changes, which it was the main task of the last code I added. MIDI converter is incomplete as well in some parts (I digged quite deep in magenta code and in its external libraries), but remember my intention is to draw generated music, which it's capped just by ProtoBuf limit.

I do agree It could all better fit into a separate `StaffSVGVisualizer` details demo page to complement class comments. I'll do it so and I'll add a link in the demo landing page.

I really appreciate your comments. I guess we both are pushing the code to improve in every iteration 🥇

rogerpasky:
I've followed all your suggestions @notwaldorf and I hope everything is ok now.

Thank you again for all your comments!

notwaldorf:
I was talking to @cghawthorne about this PR, and he mentioned vexflow [http://www.vexflow.com/], which is a library that's specifically made to draw music. Do you think this would be a better approach? And just trying to get the notes from a NoteSequence into it, rather than us adding all the drawing logic, which sounds like a nightmare to maintain for the future. WDYT?

rogerpasky:
Yep. It seems to be a great library @notwaldorf . 

I wish I had this comment on January the 9th instead spending every single night of my spare time (Saturdays and Sundays included) till 1-2 am 😔

Nevertheless, I'll dig into it and will let you know how would it get along with mm.NoteSequence

Best.

notwaldorf:
I'm really sorry about that. I didn't know about this library before, and I had sort of assumed that since you were implementing it from scratch you hadn't found a library that already did this.

rogerpasky:
I Googled it and looked into the results of the first page. VexFlow wasn't there and the rest of libs had a poor quality or a lot of overhead including a player within. None was able to mimic the existing PianoRoll visualizer (oriented to page instead to horizontal scrolling line).

As far as I've read the documentation, I'm not very confident about the complexity reduction by means of it's usage. It's very powerful indeed, but besides the double sharp and beams, I don't know if it can cover it all.

Please have in mind @notwaldorf  I'll be completely true with my final judgement. I'll be more than happy to throw away more than 200 hours of work if the result is better and the maintenance is higher. I worked on it because I really need this functionality for my own Magenta.js project, but I guess I'll spend some days studying the lib (it's quite a huge job barely understanding it).

rogerpasky:
Dear Monica,

First of all, I must apologize for this looong period of silence, but I finally reach a well formed criteria about the adoption of VexFlow as an inner component of StaffSVGVisualizer class.

## TL;DR briefing:

Doing it so is not a good idea to include VexFlow internals due to: 
* Visually declarative approach
* Very complex library
* Not documented for an easy entry

## Long explanation:

### Visually declarative approach
* It needs to be told where to put every note symbol (no note sequence interpretation). You cannot just tell "this is a MIDI 61 note (C# or Db) one quarter long starting half a quarter from the end of bar" and hope it will decide to use any accidental according to it. 
* You MUST decide if you want a flat, a sharp, a normal, or skip any accidental and decide as well to split the quarter note in two eighth notes linked by a tie you must assign to both notes.
* It has no stateful management, so previous decisions must be held by the caller to react accordingly (one previous flat in same note in current bar means no flat displaying). VexFlow does nothing about it, you must handle it (and this is the happy path, there are tonnes of exceptions which I guess I handled in minimal possible code).
* It is closer to "Finale SW" than "Logic SW", which means it is meant to draw your score instead to interpret a note sequence, just like MusicXML does (doing a MIDI to MusicXML translator is not trivial).
* I tried to find an intermediate translator (MIDI to MusicXML to VexFlow) but it was a nightmare and it would generate longer rendering times and, far much worse, greater complexity and crossed dependencies whenever any of them upgrades.
* In order to provide the NoteSequence interpretability to use VexFlow, it would mean keeping more than 80% of my code to do the protobuf content interpretation (NoteSequence) and state management to be translated to every score visual representation command. New VexFlow calling lines have not been evaluated, but I bet they will be more new lines than removed 20%
* I don't discard it in a future if needed, but nowadays it would be more a pain than a relieve.

### Very Complex Library
* 19,216 lines of code (plus dependencies) vs 1,325 in StaffSVGVisualizer.
* The number of dependencies defined in package-lock.json is 928
* VexFlow [Class Hierarchy](https://github.com/0xfe/vexflow/wiki/VexFlow-Class-Hierarchy) is pretty huge.
* Class [inner dependencies](https://github.com/0xfe/vexflow/wiki/VexFlow-Dependency-Graph) is a nightmare too.
* Actual SVG drawing code in my code is about 200 lines of code, most of it SVG path numbers, and replacing it would mean adding far much more lines calling VexFlow to cope it.
* I don't blame it for being complex. Score music representation is an extremely complex problem handling 600 years of exceptions and peculiarities. Kudos to VexFlow team, covering an excellent visualization, but it applies to another problem to the one which was intended to be solved in "NoteSequence automatic score interpretation".

### Not documented for an easy entry
* This a community requested [issue](https://github.com/0xfe/vexflow/issues/530) (_not finding the proper location for the API documentation_). 
* It has been tried to be solved one year ago with a [PR](https://github.com/0xfe/vexflow/pull/615), but it didn't progress.

## Pull Request variant proposal
I must agree my Pull Requested code is not easy (because the subject itself is not easy as well). I can understand it would be a little scary to accept this code to be maintained by the magenta/music community.

If this would be the case, given StaffSVGVisualizer is a derived class, It could be hosted in my own repo and referenced in magenta/music code to be located by other users. I would maintain this code as deeply as I would do as a contribution to magenta/music.

I've received some request in Twitter (@Seanitzel) to access my PR. I guess both choices would fit him.

My personal opinion about it is the PR should remain within magenta/music, because I guess it will be widely used, but Just let me know what do you think about it and I'll do whatever is best for the Magenta community.

notwaldorf:
Sweet, thanks for looking into that, and sorry it's been such a pain! I completely agree that your proposal of moving this code into your own library, and then having magenta/music just use that as a dependency and reference it is the best solution. That would make me far less scared of merging this, since the maintenance area for me should be pretty small.

What are you thinking, putting just the visualization code into a separate library, and calling the functions in it from the StaffSVGVisualizer? I think that would probably be easier than moving the class itself and adding a magenta dependency to your library (something something circular dependencies). Let me know what you think!

rogerpasky:
Ok, It's been a long time since last update. I had some family issues like my elder daughter university enrolment, holidays and others which were upgraded as top priorities by my wife :-D

I followed your good advice and I've created a separate library with no dependencies to Magenta.js to do the visual thing (no dependencies at all to anything, being it a plain Typescript library!!!), which will be called from StaffSVGVisualizer in a minimal code set which will be pull-requested following days.

I don't know if the "advanced demo" should remain in Magenta-js demo set, or should it stay in my own repo keeping just a link to it from a simpler demo in magenta. Please let me know that do you prefer.

notwaldorf:
I think it would make sense to keep the advanced demo in your repo (we don't, for example, demo ALL the things we can do with Tone, even though we use Tone), and have a link to it.

notwaldorf:
Also, yay, so excited to see what it looks like! Thanks so much for working on it! (and as someone who had to enroll in uni: it sucks, congrats on being done!)

rogerpasky:
I've been doing advances, and finally got it all compiling... except in demos, so I finally must give up copying the Magenta project structure and tomorrow I will follow a simpler approach.

For the record, I was trying to import @magenta/music as one of the "devDependencies", not to include in the final package but to be used in the demos, and right before end it yield something like :

```
ERROR in ./node_modules/@magenta/music/es6/index.js
Module not found: Error: Can't resolve 'tf' in '/Users/Pasky/src/staffrender/staffrender/node_modules/@magenta/music/es6'
 @ ./node_modules/@magenta/music/es6/index.js 1:82-95
 @ ./demos/demo_visualizer.ts
ℹ ｢wdm｣: Failed to compile.
```

I presume there's something wrong with my package.json because @magenta/music should be resolving internal dependencies to @tensorflow/tfjs if any, but this 'tf' thing say it is not.

Anyway, I'll start tomorrow a brand new approach. Code won't change, just structure and package creation: Given my lib is plain vanilla Typescript, I'll reduce the repo to bare bones and demo will be minimalist and native to my lib (no magenta related). Then I'll create another repo for the sake of validating the mixture with @magenta/music (importing both from it and my lib) doing a deeper and more complete demo which will break any circular reference and, if anything goes well, I'll do the Pull Request. I don't want to break anything I haven't tested properly.

I'll keep on informing.

notwaldorf:
I think your plan (of not using magenta in your package’s demo) sounds really smart!! It’s what I would do too

notwaldorf:
(And then use something like glitch or jsbin or your favourite online code thingie) to test it. Or a separate repo, whatever works for you!

rogerpasky:
I've been isolating the problem assuming this was my fault, but I guess there's something wrong with `@magenta/music/es6` due I got to reproduce it without my code (just extracting `visualizer` demo to another [repo](https://github.com/rogerpasky/magenta-viewer-demo)) and googling the problem to discover I'm not the only one with the unresolved `tf`dependency (as you can see in [issue 356](https://github.com/tensorflow/magenta-js/issues/356) you are handling as well)

Please notice my repo is using `@magenta/music` from `node_modules`, not from `../src` which could have been encapsulating the problem within the very music repo.

I documented the repo as much I could to ease the digging. Hope it helps!

rogerpasky:
Well, I think I got it. It is as simple as it can be and all the hard things related to musical interpretation (where to split the notes according to bars, etc) has been separated to another NMP module `staffrender`, which will get on improving. Documentation has been moved as well to another repo `staffrender-magentaviewer`, which shows how to create your own viewer. Links have been included in comments to appear in documentation page.

I really hope this is the final version :-)

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+   *
+   * @param sequence The `NoteSequence` to be visualized.
+   * @param div The element where the visualization should be displayed.
+   * @param config Visualization configuration options.
+   */
+  constructor(
+    sequence: INoteSequence, 
+    div: HTMLDivElement, 
+    config: StaffSVGVisualizerConfig = {}
+  ) {
```
notwaldorf: The formatting in this code is super weird. Are you using the clang plugin in vscode? (it only works if you just open the root `music` folder)

music/demos/visualizer.ts
```diff
@@ -24,11 +24,19 @@ const MIDI_URL = './melody.mid';
 
 let canvasVisualizer: mm.PianoRollCanvasVisualizer;
 let svgVisualizer: mm.PianoRollSVGVisualizer;
+let staffVisualizer: mm.StaffSVGVisualizer;
+
+let staffConfig: mm.StaffSVGVisualizerConfig = {
```
notwaldorf: can these be defaults, like we have defaults in the PianoRollVisualizer?

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
     this.drawn = false;
   }
 }
+
+/**
+ * Enumeration of different ways of horizontal score scrolling, like paginaged
```
notwaldorf: typo

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+   * 
+   * @param activeNote (Optional) If specified, this `Note` will be painted
+   * in the active color and there won't be an actual redrawing, but a 
+   * re-colouring of the involved note heads, accidentals, dots and ties 
+   * (activated and de-activated ones). Otherwise, all musical symbols which 
+   * were not processed yet will be drawn to complete the score.
+   * @param scrollIntoView (Optional) If specified and the active note to be 
+   * highlited is not visualized in the container DIV, the later will be 
+   * scrolled so that the note is viewed in the right place. This can be 
+   * altered by `AdvancedVisualizerConfig.scrollType`.
```
notwaldorf: typos: highlighted, latter,

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+   * @returns The x position of the highlighted active note relative to the 
+   * beginning of the SVG, or -1 if there wasn't any given active note. Useful
```
notwaldorf: which SVG? this class has mostly been talking about divs.

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+      this.render.scoreInfo = this.getScoreInfo(this.noteSequence);
+    }
+    const activeNoteInfo = activeNote ? 
+      this.getNoteInfo(activeNote) : undefined;
+    return this.render.redraw(activeNoteInfo, scrollIntoView);
+  }
+
+  private isNoteInInstruments(note: NoteSequence.INote) : boolean {
+    if (note.instrument === undefined || this.instruments.length === 0) {
+      return true; // No instrument information in note means no filtering
```
notwaldorf: Super nit: The Google style guide says all comments need to end in a period (here and everywhere)

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+      return this.instruments.indexOf(note.instrument) >= 0;
+    }
+  }
+
+  private timeToQuarters(time: number): number {
+    const q = time * this.noteSequence.tempos[0].qpm / 60;
+    return Math.round(q*16)/16; // Current resolution = 1/16 quarter
+  }
+
+  private quantizedStepsToQuarters(steps: number): number {
```
notwaldorf: you could combine this and the above function into one:
```
q = this.sequenceIsQuantized ? 
   value / this.noteSequence.quantizationInfo.stepsPerQuarter : 
   value * this.noteSequence.tempos[0].qpm / 60;
```

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+      this.timeToQuarters(note.endTime);
+    return {
+      start: startQ,
+      length: endQ - startQ,
+      pitch: note.pitch,
+      intensity: note.velocity
+    };
+  }
+
+  // TODO: Update when ProtocolBuffer covers a quantizedStep version
```
notwaldorf: What does this TODO mean?

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,237 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+  private getTempoInfo (
```
notwaldorf: These functions seem to only be used in one function, getScoreInfo (and some only once). I'm not sure they need to be functions at all, private or not.

rogerpasky:
Cool! I'll try to correct them all today.

I'll reply some of your comments to clarify some things about the reasoning. No worries, I'll follow all your comments.

music/src/core/visualizer.ts
```diff
@@ -441,3 +441,1266 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+      };
+    }
+    else {
+      return null;
+    }
+  }
+
+  private getPitchDetails(notePitch: number)
+  : {vSteps: number, accidental: number} {
+    const semitones = notePitch - 60;
```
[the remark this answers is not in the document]
rogerpasky: I notice this is related to the old code. Not applicable to new code but to current score resolution (1/16 of quarter note), but I guess it is not worthy to appear in `constants.ts` due it is particular to the internal module `staffrender` so I'll export it from there.

music/src/core/visualizer.ts
rogerpasky: Ok

music/src/core/visualizer.ts
rogerpasky: Ok

music/src/core/visualizer.ts
rogerpasky: Ok. Moved to DIV.

music/src/core/visualizer.ts
rogerpasky: Ok. No doubt it is super nit indeed :-D

music/src/core/visualizer.ts
rogerpasky: Nope, this has a good reason to stay as is:

If you deal just with notes, you are right, because `sequenceIsQuantized` identifies both cases (`sequence.note.startTime` and `sequence.note.quantizedStartStep`), **BUT** (and it is a huge but), protocol buffer definition of `NoteSequence` missed to keep this duality into `sequence.tempos`, `sequence.keySignatures` and `sequence.timeSignatures`because all of them hold an array of elements referenced by `time`but there's not a `quantizedStep`, so if I rely on `sequenceIsQuantized` it could fail because `time` is always meaning seconds, no matter what `sequenceIsQuantized` says.

I double-checked MIDI translation and it always fills time with seconds. That's the reason I intentionally call time or steps version in notes translation and force to time in signatures.

I guess I'm the first one who deals with tempos, key signatures and time signatures and nobody noticed it before. Music is wider than C key and 4/4.

That's re reason of the latter `TODO:` comment you mention.

music/src/core/visualizer.ts
rogerpasky: Explained before.

music/src/core/visualizer.ts
rogerpasky: Ok. It was clearer to me, because of the arrow functions in the `forEach` loop, but I can merge them in longer sentences. No problem at all.

music/src/core/visualizer.ts
rogerpasky: My fault. I've manually changed formatting to adapt to the rest of code:

* Braces: Stroustrup style with 2 spaces indentation
* Long lines: 4 spaces indentation

I tried to get the clang plugin in vscode, but there was several ones (xaver.clang-format, mitaki28.vscode-clang, kube.clangcomplete and others) and I didn't want to mess things on the rest of the file. Please, let me know which one and I'll stick to that for next contributions.

music/demos/visualizer.ts
rogerpasky: Ok

rogerpasky:
Done!

music/demos/visualizer.html
```diff
@@ -65,6 +65,13 @@ <h2>mm.PianoRollCanvasVisualizer</h2>
       <canvas id="canvas"></canvas>
     </div>
   </section>
+
+  <h2>mm.StaffSVGVisualizer</h2>
+  <section>
+    <div class="visualizer-container" id="container3">
```
notwaldorf: i don't think this ID is used anywhere

music/src/core/visualizer.ts
```diff
@@ -456,3 +458,205 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
...
+   * @param sequence The `NoteSequence` to be visualized.
+   * @param div The element where the visualization should be displayed.
+   * @param config Visualization configuration options.
+   */
+  constructor(
+      sequence: INoteSequence,
+      div: HTMLDivElement,
+      config: StaffSVGVisualizerConfig = {}) {
+    super(sequence, config);
+    if ( // Overwritting super() value. Compact visualization as default.
```
notwaldorf: I think the indentation is still whack here. It's fine if you can't fix it, I just need to know if I need to maybe do another PR to clean it up afterwards (my vscode does it automatically)

music/src/core/visualizer.ts
notwaldorf: sg, but then i would add these details to that todo (maybe just one of them, so that there's some details about it).

rogerpasky:
Thank you!!! It has been a long process, but fruitful :-)

music/demos/visualizer.html
rogerpasky: It was kept for consistency with other containers naming.

music/src/core/visualizer.ts
rogerpasky: No worries if automatic indentation rearranges it in further PRs. I'd love to have a `CONTRIBUTING.md` file in the project with all the details about the proper contribution rules, like ending comments with period or the right VSCode configuration and plug-ins. I wasn't able to configure mine to get the automation follow your rules (didn't find the proper clang plugin).

PR #285, opened 2019-03-19
Fix coconet tensor leak

notwaldorf:
Two things: 
- was leaking some tensors (no longer!)
- tslint was grumpy about not typing the tensor argument better.

I'm not super sure why the history is so whack on this PR, but I'll remove the noise when squashing.


music/src/coconet/model.ts
```diff
@@ -473,17 +473,18 @@ class Coconet {
...
       return this.getCompletionMask(pianorolls);
     } else {
-      // Create a buffer to store the input.
-      const buffer = tf.buffer([pianorolls.shape[1], 4]);
-      for (let i = 0; i < masks.length; i++) {
-        buffer.set(1, masks[i].step, masks[i].voice);
-      }
-
-      // Expand that buffer to the right shape.
-      return buffer.toTensor()
```
adarob: only this part needs to be in tidy but whatevs

notwaldorf: will change!

PR #288, opened 2019-03-21
Piano roll fixes

rogerpasky:
Several fixes added to pianoRoll visualizers:

- On top of JS float inaccuracy, redraw was removing just one single active note (added a while loop).
- The extra width space was caused by the accumulation of extra spacing inter-notes, which was changed for a subtraction in the drawing but not in the SVG / canvas general width calculation.
- Improved previous correction to make same widths on quantized and non-quantized sequences. Previous fix worked just for players using default values, but not with the ones who were giving a config.pixelsPerTimeStepvalue to the constructor.
- Horizontal paginating had a problem because of the DIV padding size which was not taken into account and ended up out of focus. I've changed it to maintain the active note in the middle of the DIV.
- Last minute correction to make the redraw() return value to be linked to the very received activeNote (wrongly  linked in polyphonic scores)

On CSS file `demos/style.css` it has been added an extra CSS value (`scroll-behavior: smooth;`) to make a more "pro" movement (yep, I know this one is not standard yet, but works perfectly in Chrome and Firefox, and doesn't break anything on Safari).


rogerpasky:
Mission accomplished @notwaldorf !!!

You can ask @adarob to unblock it whenever you want ;-)

notwaldorf:
oh, i can review it myself, don't worry about that. looking now!

music/src/core/visualizer.ts
```diff
@@ -163,22 +162,20 @@ export abstract class BaseVisualizer {
...
-        this.parentElement.scrollLeft = activeNotePosition - 20;
-      }
+      this.parentElement.scrollLeft = activeNotePosition - containerWidth * 0.5;
     }
   }
 
   protected getNoteStartTime(note: NoteSequence.INote) {
-    return this.sequenceIsQuantized ? note.quantizedStartStep : note.startTime;
+    return this.sequenceIsQuantized ? note.quantizedStartStep : 
+        Math.round(note.startTime * 100000000) / 100000000;
```
notwaldorf: Is this the same as `note.startTime.toFixed(...)` ? Either way I think this is a bit too much math to do in a loop over and over again for every note. What if you just add and subtract a small epsilon from `this.getNoteStartTime(playedNote)`, if the problem is just a floating point error? That way it's only done once per redraw.

music/src/core/visualizer.ts
```diff
@@ -365,13 +361,14 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
     }
 
     // Remove the current active note, if one exists.
-    const el = this.svg.querySelector('rect.active');
-    if (el) {
+    let el = this.svg.querySelector('rect.active');
```
notwaldorf: You can just do `let els = this.svg.querySelectorAll('rect.active')`, and just loop through that list instead.

music/src/core/visualizer.ts
rogerpasky: I tried other approaches, but they weren't that easy because you never know if epsilon must be positive or negative, therefore the need for Math.round() which is pretty efficient, and the extra * and / are very fast as well (they are bare processor primitives). I've tested it in very heavy load contexts without a problem at all.

rogerpasky: Moreover, toFixed() returns a String, not a Number.

music/src/core/visualizer.ts
rogerpasky: Done!

music/src/core/visualizer.ts
rogerpasky: Tried to find an [easier solution](https://stackoverflow.com/questions/1458633/how-to-deal-with-floating-point-number-precision-in-javascript), but they involved more maths or string parsing. I propose leaving it as is and evolve later if some other solution raises.

rogerpasky:
@notwaldorf I'm afraid I'm going to bed (it's 1am here in Spain). I think both change requests should be closed: 

- First one is already pretty efficient indeed (just rounding numbers, not tensor manipulation :-D)
- Second one already refactored.

See you tomorrow (virtually).

music/src/core/visualizer.ts
rogerpasky: @notwaldorf I've done some extra-latency testings with solid and consistent results:

```typescript
  protected getNoteStartTime(note: NoteSequence.INote) {
    return this.sequenceIsQuantized ? note.quantizedStartStep : 
        Math.round(note.startTime * 100000000) / 100000000;
  }

  protected getNoteStartTimeOld(note: NoteSequence.INote) {
    return this.sequenceIsQuantized ? note.quantizedStartStep : note.startTime;
  }

  protected testRounding(n: number): number {
    // Data setup
    let notes = new Array<NoteSequence.INote>(n);
    for (let i = 0; i < n; ++i) {
      const note = new NoteSequence.Note;
      note.startTime = 1/i;
      notes[i] = note;
    }

    // First roundup
    let timeOldBegin = (new Date()).getTime();
    for (let i = 0; i < n; ++i) { this.getNoteStartTimeOld(notes[i]); }
    let timeOldEnd = (new Date()).getTime();
    let timeNewBegin = (new Date()).getTime();
    for (let i = 0; i < n; ++i) { this.getNoteStartTime(notes[i]); }
    let timeNewEnd = (new Date()).getTime();

    // Second roundup to avoid data caching effect
    timeOldBegin = (new Date()).getTime();
    for (let i = 0; i < n; ++i) { this.getNoteStartTimeOld(notes[i]); }
    timeOldEnd = (new Date()).getTime();
    timeNewBegin = (new Date()).getTime();
    for (let i = 0; i < n; ++i) { this.getNoteStartTime(notes[i]); }
    timeNewEnd = (new Date()).getTime();
    
    console.log('oldPost - oldPre  = ', timeOldEnd - timeOldBegin);
    console.log('newPost - newPre  = ', timeNewEnd - timeNewBegin);
    console.log('Latency per request = ', 
      ((timeNewEnd-timeNewBegin) - (timeOldEnd - timeOldBegin)) / n);
    return (timeNewEnd-timeNewBegin) - (timeOldEnd - timeOldBegin);
  }
```
Running it in Chrome console over an instance having `this.sequenceIsQuantized == null`(which is where the new quantization applies) yields following results:
```console
> this.testRounding(10000000);
oldPost - oldPre  =  477
newPost - newPre  =  525
Latency per request =  0.0000048
48
>
```
I guess five millionth of millisecond (4.8 nanosecond) per call is good enough latency to let it go. Please trust in wonderful V8 JavaScript JIT compiler magic and accept the PR :-).

notwaldorf: sg, no worries

music/src/core/visualizer.ts
```diff
@@ -163,22 +162,20 @@ export abstract class BaseVisualizer {
...
       scrollIntoView: boolean, activeNotePosition: number) {
-    if (scrollIntoView && this.parentElement) {
-      // See if we need to scroll the container.
+    if (scrollIntoView) {
       const containerWidth = this.parentElement.getBoundingClientRect().width;
-      if (activeNotePosition >
-          (this.parentElement.scrollLeft + containerWidth)) {
-        this.parentElement.scrollLeft = activeNotePosition - 20;
-      }
+      this.parentElement.scrollLeft = activeNotePosition - containerWidth * 0.5;
```
notwaldorf: Let's not change this, please. You can overwrite this behaviour if you want in a custom visualizer, but I find the "keep the active note in the middle" a worse experience (it moves a lot too much and it makes it harder to follow what's happening).

rogerpasky: Sure! Will do. 

Please notice current pagination scrolling has a cummulative bug due to extra width of the outer DIV padding (noticeable in longer scores). Not very urgent stuff, so I'll try to fix it in further contributions.

rogerpasky: @notwaldorf Done!

rogerpasky:
Is there needed any further change @notwaldorf ?

notwaldorf:
This is great, thanks so much again!

rogerpasky:
You are welcome!

I keep on testing very hard the multi-key multi time-signature. Looks promising, but simplifying 600 years of score rules in a coherent subset is not that easy... I'll let you know when is Ok to start reviewing the code in the other PR.

notwaldorf:
I did a very small pass (just by looking at the demo), and left some comments

Issue #281, opened 2019-03-14
How to use svg visualizer


kadinzhang:
I saw that #260 was pushed out to resolve some issues with the canvas visualizer. At the time of me writing this, [docs](https://tensorflow.github.io/magenta-js/music/classes/_core_visualizer_.visualizer.html) haven't been updated - how can I use this mm.PianoRollSVGVisualizer? I tried using the same params as the old Visualizer(), swapping out the canvas element with a generic div, and although it returned no errors, nothing was rendered.

notwaldorf:
Sorry about that! We've updated the [docs](https://tensorflow.github.io/magenta-js/music/classes/_core_visualizer_.pianorollsvgvisualizer.html) -- the parameter to the constructor is an `svg` element, not a `div`. Here's a [usage example](https://github.com/tensorflow/magenta-js/blob/master/music/demos/visualizer.ts#L74).

We should probably throw an error if the tag name of the thing you pass to the constructor isn't `svg`, since that would've sorted out your question. Would you be interested in sending that PR?

kadinzhang:
Sure! That sounds fun. I'll give it a try.

notwaldorf:
This was fixed in https://github.com/tensorflow/magenta-js/pull/291

PR #291, opened 2019-03-30
Add warning if element passed to PianoRollSVGVisualizer isn't svg #281

kadinzhang:
Please let me know if this isn't the correct way to do it - I'm still learning typescript.

### DIFF
--- a/music/src/core/visualizer.ts
+++ b/music/src/core/visualizer.ts
@@ -330,7 +330,13 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
       sequence: INoteSequence, svg: SVGSVGElement,
       config: VisualizerConfig = {}) {
     super(sequence, config);
-
+        
+    //Return error if element isn't svg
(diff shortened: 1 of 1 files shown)


kadinzhang:
I signed it

music/src/core/visualizer.ts
```diff
@@ -330,7 +330,13 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
       sequence: INoteSequence, svg: SVGSVGElement,
       config: VisualizerConfig = {}) {
     super(sequence, config);
-
+        
+    //Return error if element isn't svg
+    if (!(svg instanceof SVGSVGElement)) {
```
notwaldorf: What do you think about being a bit more aggressive and throwing an Error instead? So something like `throw new Error('mm.PianoRollSVGVisualizer requires an <svg> element to display the visualization');`

kadinzhang: Sure!

PR #292, opened 2019-04-11
Hacky patch for chord-conditioned Piano Genie model

chrisdonahue:
As discussed

### DIFF
--- a/music/src/piano_genie/model.ts
+++ b/music/src/piano_genie/model.ts
@@ -63,7 +63,7 @@ function disposeState(state: LSTMState) {
 /**
  * Root note of chord for chord-conditioned model.
  */
-enum ChordRoot {
+enum PitchClass {
   None = 0,
   C,
   Cs,
@@ -94,7 +94,8 @@ enum ChordFamily {
   Min7b5
 }
 
-type Chord = [ChordRoot, ChordFamily];
+type Chord = {root: PitchClass, family: ChordFamily};
+const NOCHORD: Chord = {root: PitchClass.None, family: ChordFamily.None};
 
 /**
  * Samples logits with temperature.
@@ -211,7 +212,7 @@ class PianoGenie {
     // This runs the model once to force Tensorflow JS to allocate necessary
     // memory. Otherwise the prediction will take a long time when the user
     // presses a button for the first time.
-    this.next(0);
+    this.next(0, NOCHORD);
     this.resetState();
   }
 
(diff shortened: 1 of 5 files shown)


music/src/piano_genie/model.ts
```diff
@@ -2,7 +2,7 @@
  * Core implementation for Piano Genie model.
  *
  * @license
- * Copyright 2018 Google Inc. All Rights Reserved.
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
notwaldorf: this doesn't actually need to change btw :)

music/src/piano_genie/model.ts
```diff
@@ -60,6 +60,43 @@ function disposeState(state: LSTMState) {
   }
 }
 
+/**
+ * Root note of chord for chord-conditioned model.
+ */
+enum PitchClass {
```
notwaldorf: Can we reuse https://github.com/tensorflow/magenta-js/blob/master/music/src/protobuf/proto.d.ts#L390?

music/src/piano_genie/model_test.ts
```diff
@@ -64,17 +64,17 @@ test('Piano Genie Model Correctness', async (t: test.Test) => {
...
     for (let i = 7; i >= 0; --i) {
       genie.overrideDeltaTime(0.1);
-      keys.push(genie.next(i, 0.));
+      keys.push(genie.next(i, NOCHORD, 0.));
     }
     // Fast trill with temperature 0.5.
     for (let i = 0; i < 8; ++i) {
       genie.overrideDeltaTime(0.05);
-      keys.push(genie.next(3 + (i % 2), 0.5, 1337));
+      keys.push(genie.next(3 + (i % 2), NOCHORD, 0.5, 1337));
```
notwaldorf: i think NOCHORD should be the default, and it should be the _last_ argument, so that this isn't a breaking change

notwaldorf: (so that you only have to use it if you don't want the default)

music/src/piano_genie/model.ts
```diff
@@ -186,14 +223,15 @@ class PianoGenie {
    * application in real time (it keeps track of time internally).
    *
    * @param button Button number (one of {0, 1, 2, 3, 4, 5, 6, 7}).
+   * @param chord Chord type.
    * @param temperature Temperature. From 0 to 1, goes from argmax to random.
    * @param seed Random seed. Use a fixed number to get reproducible output.
    */
-  next(button: number, temperature?: number, seed?: number) {
+  next(button: number, chord: Chord, temperature?: number, seed?: number) {
```
notwaldorf: (as per below comment), i think this should be `chord=NOCHORD` as a default, and moved to the end

music/src/piano_genie/model.ts
chrisdonahue: Thanks! That's good to know :)

chrisdonahue:
Sorry I should have added some context on this (have a chat going with Adam). This patch is far from ready for merging. It breaks the old Piano Genie model by hard coding the new chord conditioning requirements. I was told to open the PR prematurely to get things going on an application. Will do a proper PR next week after ISMIR. I intend to make a new abstract parent class and subclass it to the original `PianoGenie` and `PianoGenieChordConditioned`.

adarob:
Let's just close this for now and I can look at your fork.

PR #305, opened 2019-05-02
Treat time of undefined as 0 when converting ns to midi.

### LINKED ISSUE #302: Groove returns a note sequence that doesn't have tempos set correctly
It seems that if I do:
```
const recon = await grooveModel.decode(z, temp, undefined, 4, someTempo);
const m = mm.sequenceProtoToMidi(recon[0])
```
This throws a `Uncaught Error: NoteSequence must have exactly 1 tempo at time 0"` on the last line. We probably should set the tempos correctly.

adarob:
Fixes #302

### DIFF
--- a/music/src/core/midi_io.ts
+++ b/music/src/core/midi_io.ts
@@ -129,12 +129,12 @@ export function sequenceProtoToMidi(ns: INoteSequence) {
     ns.timeSignatures = [{time: 0, numerator: 4, denominator: 4}];
   }
 
-  if (ns.tempos.length !== 1 || isZeroOrUndefined(ns.tempos[0].time)) {
+  if (ns.tempos.length !== 1 || !isZeroOrUndefined(ns.tempos[0].time)) {
     throw new MidiConversionError(
         'NoteSequence must have exactly 1 tempo at time 0');
   }
   if (ns.timeSignatures.length !== 1 ||
-      isZeroOrUndefined(ns.timeSignatures[0].time)) {
+      !isZeroOrUndefined(ns.timeSignatures[0].time)) {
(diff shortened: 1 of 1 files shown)


music/src/core/midi_io.ts
```diff
@@ -120,18 +120,21 @@ export function sequenceProtoToMidi(ns: INoteSequence) {
...
+
   if (!ns.tempos || ns.tempos.length === 0) {
     ns.tempos = [{time: 0, qpm: constants.DEFAULT_QUARTERS_PER_MINUTE}];
   }
   if (!ns.timeSignatures || ns.timeSignatures.length === 0) {
     ns.timeSignatures = [{time: 0, numerator: 4, denominator: 4}];
   }
 
-  if (ns.tempos.length !== 1 || ns.tempos[0].time !== 0) {
+  if (ns.tempos.length !== 1 || isZeroOrUndefined(ns.tempos[0].time)) {
```
notwaldorf: Don’t you want the negation here, since it said !== 0 originally?

music/src/core/midi_io.ts
```diff
@@ -120,18 +120,21 @@ export function sequenceProtoToMidi(ns: INoteSequence) {
...
     throw new MidiConversionError(
         'NoteSequence must have exactly 1 tempo at time 0');
   }
-  if (ns.timeSignatures.length !== 1 || ns.timeSignatures[0].time !== 0) {
+  if (ns.timeSignatures.length !== 1 ||
+      isZeroOrUndefined(ns.timeSignatures[0].time)) {
```
notwaldorf: Here too

PR #307, opened 2019-05-06
documentation improvement, resolves #282

### LINKED ISSUE #282: How to use MusicRNN continuesequence
In the [docs](https://tensorflow.github.io/magenta-js/music/classes/_music_rnn_model_.musicrnn.html#continuesequenceandreturnprobabilities), it gives the parameters for continuesequence() as:

 `continueSequence(sequence: INoteSequence, steps: number, temperature?: number, chordProgression?: string[]): Promise<INoteSequence>`. 

I'm confused about the temperature and chordProgression parameters, as there isn't much elaboration in the documentation. For temperature, what is the range of numbers that can be imputed? I tried a variety of numbers, and all but 1 caused the method to return nothing back. For chordProgression, how would I format the progression? Is there an example I can see that demonstrates how to use it?

Thanks!

mattetti:
As per @notwaldorf suggestion, here is a PR adding some documentation on the format of the optional params of `continueSequence`. Resolves #282 
I added a link to the Tonal package's GitHub since I thought it was better than the npm link but let me know if that's not correct.
I also tested the documentation generation but noticed in the history of this project that the documentation is only generated when we reach a release milestone so I didn't commit the generated doc changes.

### DIFF
--- a/music/src/music_rnn/model.ts
+++ b/music/src/music_rnn/model.ts
@@ -200,9 +200,13 @@ export class MusicRNN {
    * @param sequence The sequence to continue. Must be quantized.
    * @param steps How many steps to continue.
    * @param temperature (Optional) The softmax temperature to use when sampling
-   * from the logits. Argmax is used if not provided.
+   * from the logits. Argmax is used if not provided. Temperature can be any
+   * number value above 0, however, anything above 1.5 will essentially result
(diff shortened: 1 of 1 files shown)


notwaldorf:
Looks great. Thanks so much!

PR #313, opened 2019-05-14
[WIP] Update magenta/image to tfjs 1.x

### LINKED ISSUE #296: Upgrade @magenta/image to tfjs 1.0.x
Hi,

I was trying to use the [BodyPix](https://github.com/tensorflow/tfjs-models/tree/master/body-pix) model with Magenta but the tfjs version required is 1.0.0. Is it possible to update it ?

Thanks!

notwaldorf:
Fixes #296.

- updates to tfjs 1.x
- moves from browserify to webpack

Note: Does not work right now because the `loadFrozenModel` -> `loadGraphModel` change throws an error:
```
tf-core.esm.js:17 Uncaught (in promise) Error: The JSON from HTTP path https://storage.googleapis.com/magentadata/js/checkpoints/style/arbitrary/predictor/weights_manifest.json contains neither model topology or manifest for weights.
    at t.<anonymous> (tf-core.esm.js:17)
    at tf-core.esm.js:17
    at Object.next (tf-core.esm.js:17)
    at i (tf-core.esm.js:17)
```

cc @vdumoulin


notwaldorf:
@vdumoulin it looks like based on https://github.com/tensorflow/tfjs/issues/285#issuecomment-388086960 we need to update the model weights to also contain the architecture?

cc @nsthorat to confirm

notwaldorf:
update: we gotta convert this model with the new 1.0 converter. cc @adarob @vdumoulin

vdumoulin:
@reiinakano do you still have the model checkpoint lying around?

reiinakano:
I updated my demo awhile ago to tfjs v1.0.4. https://github.com/reiinakano/arbitrary-image-stylization-tfjs/pull/23

Feel free to use the new checkpoints I uploaded in the same PR.

Note that it will be much slower unless you do `tf.ENV.set('WEBGL_PACK', false);` as discussed in https://github.com/tensorflow/tfjs/issues/1528

tafsiri:
Hi @notwaldorf @vdumoulin,

I was wondering if you've been able to make any headway on this. I've gotten a version of the code working with tfjs 1.2.x with the converted checkpoints that @reiinakano  linked to above.

However I did notice that the model doesn’t work on iOS (blank outputs). I think it's underflowing because of the fp16 issue mentioned in [the blog post](https://magenta.tensorflow.org/blog/2018/12/20/style-transfer-js/). I imagine the same weight scaling approach would need to be done on @ updated conversion (or maybe re-converting a version you have that has that scaling applied? If the scaling code is available I’d be happy to give it a go.

Secondly I was wondering if you would be interested in a PR that adds an inference function that take and return tensors (as opposed to images)? We’ve found adding these ‘lower level’ functions useful in some of the tensorflow.js models to allow them to work outside of the browser (they avoid calls to browser specific image transforms). The demo I was working on using this was outside of the browser.

reiinakano:
That's odd. I'm pretty sure I just converted the checkpoints with scaling already applied using tfjs-converter

tafsiri:
@reiinakano I tried loading up the demo on your site in Safari (iphone X) and got an all black image. This was with the distilled mobilenet style model and the separable conv2d transformer. Let me know if i missed anything. Maybe its something else?

reiinakano:
Does it fail for the original transfer network too? I definitely remember scaling both, but never tried testing the separable conv 2d one since I don't have access to osx.

 For the original transfer network, I'm pretty sure this is the same scaled one in magenta right now which I just converted directly.

tafsiri:
@reiinakano Thanks! you are right! it works with the original transformer network. Just not the separable one.

Issue #309, opened 2019-05-06
update tone.js dependency


mattetti:
A few issues are related to the fact that this project relies on a old version of tone.js which was since improved and bugs/dependencies fixed. https://github.com/Tonejs/Tone.js/blob/c023181579d55bac110b9d369f79aa5acd275499/CHANGELOG.md#1349

#56 #294 should all be fixed if we can update tone.js @notwaldorf mentioned issues trying to update so it might be trickier than it sounds.

notwaldorf:
Ok, so I tried this again, and wanted to leave some more details, for posterity. I tried updating to `13.4.9`. If you build the demos and go to `player.html`, this is what I see:
- trying to play a note sequence starts somewhere off (like, at the 3rd note, not the first one). Stopping the player at this point doesn't always completely stop it, and sometimes leaves the last note that was playing held forever (and sounds gnarly)
- in the case where that _didn't_ happen and the player stopped correctly, starting the note sequence again will pay it correctly (from the first note, etc).

It almost looks like we're trying to play it before the Tone transport is ready, but I'm not really sure how to debug this (and fewer ideas about how to fix this).

cc @tambien

tambien:
i think my suggestion would be to try to decouple the players from the Transport. it should be possible to use Tone.Clock in a similar way as you're currently adding things to the Transport, but it would require a bit of reworking. Also for shorter segments, you could just schedule all of the notes upfront instead of playing them through the Tone.Part callback (for sections that are a few minutes long, this sometimes gets too congested to play easily). Then to stop it, you can just disconnect and dispose the player object which would throw away all future events. 

i can take a look at this when i have a moment, if no one else gets to it before me.

notwaldorf:
Also, on that demo page I still see 260+ `AudioContext` warnings in Chrome, that come from inside of Tone.js (it looks like Tone is very aggressively trying to start the AudioContext without any user interaction, which doesn't fly. It's not a problem, it's just a very spammy console warning, i guess

notwaldorf:
@tambien I'd loooove if you could take a look instead of me because I'm 100% going to break this.

mattetti:
I took a quick uneducated look and while I don't see the `AudioContext` errors going to http://localhost:8080/player.html using the latest tone.js, I do noticed that the base player sequence starts off and there are a lot of clicks. The sequence once played out is fine again but the clicks are still there. I decided to download the MIDI and surely enough the sequence conversion is off now:

[...]

Issue #56, opened 2018-04-26
Remove Tone.js from JavaScript package? 


teropa:
The package at https://cdn.jsdelivr.net/npm/@magenta/music@0.0.5/dist/magentamusic.min.js now includes Tone.js as well, as the player using it is exported from `core.ts`.

I don't think this is something users would expect, and may cause issues with projects using other versions of Tone.

adarob:
I like having the player accessible from our package. Is there any way to
make this work?

adarob:
@cghawthorne

teropa:
Should still be possible by using an NPM peer dependency, and possibly
configuring Browserify so that it doesn't include Tone but just expects it
to be around.

2018-04-26 5:27 GMT+00:00 Adam Roberts <notifications@github.com>:

> @cghawthorne
>
>

adarob:
But that would make this example no longer work, correct?
https://github.com/tensorflow/magenta-js/blob/master/music/README.md#getting-started

adarob:
Yeah, I think we'll need something like gulp to do that:
https://9elements.com/io/external-bundles-with-browserify-and-gulp/

adarob:
@tambien do you have a suggestion for handling this?

tambien:
i'm actually working on this myself. Tone.js currently doesn't play too nice when it's loaded multiple times on a page. i was talking to someone yesterday about this and may have a solution to the problem. i think it could be as simple as setting a browser global so that all versions can access the same AudioContext.

adarob:
I was also looking into leaving it out of our bundle and expecting it to
loaded separately if our player is used.  However it seems that tone is not
found by my bundle even if I load the tone bundle first. If I browserify my
own bundle with only tone in it and load that first, it does work. Any
ideas?

adarob:
I think I know how to do it now. I'll give it a try this afternoon.

tambien:
@teropa i've got a question, is the reason that you don't want Tone.js packaged into magenta is bc you're using Tone.js in the rest of your app and that'll cause two versions of Tone.js to be loaded? 

talking with @adarob about this, i think i can add a global to Tone.js to make it so that multiple instances running at the same time share the same AudioContext and so they play nice with each other. it will obviously add some to the package size and also log **Tone.js rX** multiple times in the console... but i'm thinking that might be better for versioning and ease of install than a peerDep.

teropa:
[...]

PR #316, opened 2019-05-23
Upgrade Tone.js 

tambien:
I've updated to the latest Tone.js. It caused a glitch with the PolySynth which seemed to be because the synths were created right before they were used which didn't give it enough time to initialize the synths. Instead i've created the synths in the constructor and then reference those synths in the `getSynth` method. The only change is that might come up is that you could run out of synth voices if you were counting on each of the instruments actually needed 10 note polyphony. in this case the PolySynth constructor could just be made with a larger number in the constructor. 

Addresses #309 #56


tambien:
signed it!

notwaldorf:
Looks good, I think you have a lint error according to Travis :)

notwaldorf:
so i pulled this locally and everything mostly sounds ok, but i hear like...a pop on basically every second note (it's light and overlapped with the note, as if it were a click). can this make sense? it's just the basic player demo, and then i can repro this on basically every non-soundfont, without having the extra click track enabled.

mattetti:
I noticed the pop in my exploration and it was due to the not being repeated for a short amount of time right at the end of the expected length (see the piano roll screenshot I added to the issue). Unfortunately, I'm not sure what's causing that

mattetti:
I took a closer look and this time the MIDI is totally fine however I do hear the unexpected clicks. I recorded the PR preview and the web for comparison:

Tone 13.8.12.mov: https://drive.google.com/file/d/1r4SLi1R1i9HyYGfHUfRocjbKTo0GBn4X/view?usp=sharing (running locally)

Tone r12 (currently used version): https://drive.google.com/file/d/11nnWxgkkzKRc4qKNKnD5uE93u7aYVHAx/view?usp=sharing  using https://tensorflow.github.io/magenta-js/music/demos/player.html

mattetti:
To be clear, I think the issue is with Tone's engine that might be clicking so my recommendation is to accept this PR so we are on the latest version of tone, even if tone might have a smoothing issue switching from one note to another which might lead to a click.

notwaldorf:
Updating this with some offline conversation: even if the problem is Tone related, we can’t merge this PR, since this will make all the synth players in that version of Magenta unusable, and I’m not comfortable tagging that release. Since 100% of the reason why we have a Player class is to allow people to play audio easily and smoothly, any change that makes these players not work as well as before is a bad change, in my opinion.

mattetti:
That makes total sense, I'm investigating this week. Starting by checking across browsers to see if the issue is Chrome related or consistent across all major browsers.

mattetti:
macOS 10.14.5 (18F132)
Tone.js v13.4.9
mm.Player test: http://localhost:8080/player.html

Repro steps:
1. load the player page
2. press the first play button in the Controls section
Expected: notes, play one at a time, with the playing note decaying quickly or stopping when the next note plays.

3. stop the playback and press play again
Expected: playback is similar to the first time, one note at a time, starting from the first note

4. Scroll to the Changing the tempo section and press play
Expected: Musical sequence plays one note at a time, starting from the first note.

5. Reload the player test page, re-do step 4
Expected: everything plays as expected.

6. Adjust the tempo while playing
Expected: the music notes are triggered faster or slower based on the set tempo.
----

Firefox 68.0 (64-bit) - Expectations 2, 4, 5 fail with the first note often getting stuck (keeps on playing for longer than expected and some audio cracks can be heard later on). Expectation 3 seems to pass most of the time (in other words, the playback works the second time around). Some notes get stuck never getting released, requiring a page reload to stop the sine wave sound. Repro step, start the Controls sequence and press stop before the 7th note (when the first note turns off)

Safari Version 12.1.1 (14607.2.6.1.1) - similar behavior as FF. 

Chrome 75.0.3770.100 - same behavior as FF.

----

There is clearly an issue with the first note being triggered, it might be a race condition or an initialization bug the problem happens at the beginning of the sequence. Once the sequence played, following tentatives result in a positive outcome. This seems to indicate some sort of initialization/state problem. 
Unrelated (IMHO), the console is reporting and issue when the player page loads:
`Unhandled Promise Rejection: TypeError: null is not an object (evaluating 'outputs.map')` @ player.ts 101

```typescript
function setupMIDIPlayerDemo() {
  const playBtn = document.getElementById('midi-play') as HTMLButtonElement;
  const midiOutputs =
      document.getElementById('midi-outputs') as HTMLSelectElement;
  const player = new mm.MIDIPlayer();

  player.requestMIDIAccess().then((outputs: WebMidi.MIDIOutput[]) => {
    midiOutputs.innerHTML =
        outputs.map(device => `<option>${device.name}</option>`).join('');  // <---
    playBtn.disabled = outputs.length === 0;
  });
```

This is a red herring, fixing the code by adding a null check doesn't change the behavior.

notwaldorf:
Yeah, `setupMIDIPlayerDemo` shouldn't have anything to do with the mm.Player, that bit of the demo is looking for connected MIDI devices, and you probably don't have one. A null check will fix that (`midiOutputs.innerHTML = outputs ? ... : ''`. or something along those lines).

mattetti:
Debugging notes for myself and potentially others:

Our entry point is in `src/core/player.ts` where the player's start method is defined:

```ts
    this.resumeContext();
    const isQuantized = sequences.isQuantizedSequence(seq);
    if (this.playClick && isQuantized) {
      seq = this.makeClickSequence(seq);
    }
    if (qpm) {
      Tone.Transport.bpm.value = qpm;
    } else if (seq.tempos && seq.tempos.length > 0 && seq.tempos[0].qpm > 0) {
      Tone.Transport.bpm.value = seq.tempos[0].qpm;
    } else {
      Tone.Transport.bpm.value = constants.DEFAULT_QUARTERS_PER_MINUTE;
    }
    if (isQuantized) {
      seq = sequences.unquantizeSequence(seq, qpm);
    } else if (qpm) {
      throw new Error('Cannot specify a `qpm` for a non-quantized sequence.');
    }

    this.currentPart = new Tone.Part((t: number, n: NoteSequence.INote) => {
      if (this.playClick ||
          (n.pitch !== constants.LO_CLICK_PITCH &&
           n.pitch !== constants.HI_CLICK_PITCH)) {
        this.playNote(t, n);
      }
      if (this.callbackObject) {
        Tone.Draw.schedule(() => {
          this.callbackObject.run(n, t);
        }, t);
      }
    }, seq.notes.map(n => [n.startTime, n]));
    if (this.desiredQPM) {
      Tone.Transport.bpm.value = this.desiredQPM;
    }
    this.currentPart.start();
    if (Tone.Transport.state !== 'started') {
      Tone.Transport.start();
    }
    return new Promise(resolve => {
      this.scheduledStop = Tone.Transport.schedule(() => {
        this.stop();
        resolve();
        if (this.callbackObject) {
          this.callbackObject.stop();
        }
      }, `+${seq.totalTime}`);
    });
  }
```

We are using [Tone's Transport class](https://tonejs.github.io/docs/r13/Transport) to handle  scheduled events.

Here's a reduction of the code being executed:

```ts
Tone.Transport.bpm.value = 60;
this.currentPart = new Tone.Part((t: number, n: NoteSequence.INote) => {
        this.playNote(t, n);
}, seq.notes.map(n => [n.startTime, n]));
this.currentPart.start();
Tone.Transport.start();
```
So, we set the transport's tempo and create a [Tone's Part](https://tonejs.github.io/docs/r13/Part) which is a translation of our passed note note sequence. The [part is then started](https://tonejs.github.io/docs/r13/Part#start) and then the Transport is started.

So the interesting part is that we start the part before the transport. Running a quick experiment, I delayed the start of the part:
```ts
this.currentPart.start(0, 4);
```
And the "stuck note" is gone, but we hear clicks at note change. If my assumption is correct, it seems like the issue might be related to how we start the part and the transport and the time difference between the 2.

mattetti:
Went to dig further and looked into the `this.playNote(t,n)` function call since that's implemented in our player.

The implementation looks like that:

```ts
export class Player extends BasePlayer {
  private synths = new Map<number, any>();  // tslint:disable-line:no-any
  private drumKit = DrumKit.getInstance();

  /**
   * The Tone module being used.
   */
  static readonly tone = Tone;  // tslint:disable-line:no-any

  protected playNote(time: number, note: NoteSequence.INote) {
    // If there's a velocity, use it.
    const velocity = note.hasOwnProperty('velocity') ?
        note.velocity / constants.MAX_MIDI_VELOCITY :
        undefined;

    if (note.isDrum) {
      this.drumKit.playNote(note.pitch, time, velocity);
    } else {
      const freq = new Tone.Frequency(note.pitch, 'midi');
      const dur = note.endTime - note.startTime;
      this.getSynth(note.instrument, note.program)
          .triggerAttackRelease(freq, dur, time, velocity);
    }
  }
```
The first note looks like this: `{"pitch":60,"startTime":0,"endTime":0.5,"program":0,"quantizedStartStep":"0","quantizedEndStep":"2"}`
So the executed code like this:

```ts
protected playNote(time: number, note: NoteSequence.INote) {
      const freq = new Tone.Frequency(note.pitch, 'midi');
      const dur = note.endTime - note.startTime;
      this.getSynth(note.instrument, note.program)
          .triggerAttackRelease(freq, dur, time, velocity);
  }
```

We need to take a look into `getSynth`:

```ts
private getSynth(instrument: number, program?: number) {
    if (this.synths.has(instrument)) {
      return this.synths.get(instrument);
    } else if (program !== undefined && program >= 32 && program <= 39) {
      const bass = new Tone.Synth({oscillator: {type: 'triangle'}}).toMaster();
      bass.volume.value = 5;
      this.synths.set(instrument, bass);
    } else {
      this.synths.set(instrument, new Tone.PolySynth(10).toMaster());
    }
    return this.synths.get(instrument);
  }
```

That's the code @tambien changed in this PR. It seems very likely that the problem we are facing is located in this part of the code.

The first time around, `this.synths` is empty, so we don't hit the cache and instead creating a new synth instance and fetching it from the cache:

```ts
this.synths.set(instrument, new Tone.PolySynth(10).toMaster());
return this.synths.get(instrument);
```

To prove my theory that the issue is with the polysynth initialization happening when we try to start the sequence, I tried to warm up the cache by setting the synth when the player is created:

```ts
export class Player extends BasePlayer {
  private synths = new Map<number, any>();  // tslint:disable-line:no-any
  private drumKit = DrumKit.getInstance();

  constructor(playClick = false, callbackObject?: BasePlayerCallback) {
    super(playClick, callbackObject);
    this.synths.set(0, new Tone.PolySynth(10).toMaster());
  }
  //...
}
```
This helped resolved part of the problem, however because we are also creating the player on the fly and because the polysynth creation is really slow, it doesn't totally solve the problem.

@notwaldorf I agree that this is a much bigger than I initially thought. It's also problematic because it's due to the way we use Tone.js so it's not intrinsically a Tone.js issue, however it's a regression indeed.

My naive first thought is that we should never define core elements of the audio processing chain on the fly when starting playing a sequence. There is a high chance of skipped frames or audio glitches when initializing audio generators or processors on the fly. I realize it would makes the API player more cumbersome and not backwards compatible with but I'd suggest that we add a blocking setup stage to the API. Playback should only be initiated once the audio chain is in place and ready, probably via a callback. It's a pretty major API change but I think it might be a good opportunity to make player instrument setup less implicit allowing the developer to setup a drumkit and or a synth. Ideally, the drumkit would become an interface so developers could bring their own drumkit (such as a sample based drumkit and custom mappings).

tambien:
The clicking is potentially caused by a really fast attack of the PolySynth. The new PolySynth implementation struggles to allocate a new voice and generate a really fast attack envelope in this situation. 

The quick fix is just to make [the envelope a little slower](https://github.com/tambien/magenta-js/blob/master/music/src/core/player.ts#L380). 

To give a little more context, the major change between v12 and v13 in the PolySynth implementation is that the previously all of the synth oscillators were created ahead of time and managed statically. This is helpful to reduce the performance overhead of potentially having a lot of oscillators running silently. The downside is as @mattetti mentioned, there is a cost to initializing Web Audio components. So it seems like in this situation it struggles to initialize a new oscillator and envelope and performance a fast attack. 

If we were to do a refactor of the mm.Player, i might try to decouple it from the Transport altogether. On the demo page you can get into a few weird situations if you start a new sequence before another one has finished, or don't explicitly hit "stop" which might cause multiple players to all be playing at once because they are all synced to the same transport. 

I might suggest ditching the transport and just scheduling all notes up front when `start` is invoked. It would look something like this: 

```javascript
start(seq){
  const synth = new Tone.PolySynth(10).toMaster()
  const now = Tone.now()
  seq.notes.forEach(note => {
    const freq = new Tone.Frequency(note.pitch, 'midi');
    synth.triggerAttackRelease(freq, note.endTime - note.startTime, now + note.startTime, note.velocity)
  })
  //set a timeout for after the last note has stopped and dispose of the synth. 
}
```

The things that you lose with this implementation is the ability to set the tempo dynamically while the player is playing. as well as the ability to pause and resume a player. `stop` could be implemented by simply connecting the synth to a gainNode instead of directly to the master output and then just disconnect the gain node to stop the sound (maybe after a short ramp to avoid clicks). The upside would be a much simpler API which doesn't rely on as many complex components. 

I could mock something up if that's helpful in another branch, but the API would be subtly different than the current one.

mattetti:
@cannoneyed suggested a temporary workaround that I find quite interesting as a stop gap solution: https://github.com/tensorflow/magenta-js/pull/325

The challenge is that:
* it doesn't solve the overall issue due to the change of behavior in tone.js that now creates the situation we are in.
* doesn't allow to have more than 1 instrument per player (bass excluded) so you can't make a sequencer for instance.

I think someone on the project should make a decision on how we want to move forward as we have many options, none of them ideal but all doable.

cannoneyed:
@mattetti @notwaldorf I've got an idea that I'm going to try right now - basically, we can defer initialization of the voices for a sequence by introducing a slight delay before initial playback - when the `start` method is called, we'll go through each voice in the sequence to be played an create the correct PolySynth before we start actual playback (plus initialize a bass, regardless). This will be mostly transparent to the user (there might be a tiny delay when pressing play, depending on how long it takes to initialize) and it should fix the glitch issues

mattetti:
I like the suggestion, but how will you know that everything is initialized? Those are non blocking calls so the behavior might be uncertain.

cannoneyed:
Right now i'm looking into the behavior and seeing how it all behaves.... seems like naively implementing the pre-start initialization gets rid of the glitches at the start of playback without any special treatment for the non-blockingness, but the little tiny cracks and pops are still there no matter what I try envelope-wise... seems like there's a bit more to figure out here!

tambien:
I did more research into that popping sound and there seems to be a bug in Chromium which causes a discontinuity in some specific cases when you schedule ramps. 

This demo mimics the kind of curves which are causing the pops: https://codesandbox.io/s/silly-fog-xh9sl

This is the output in Chrome: 
image

and this is the output in Firefox: 
image

Notice that big discontinuity in Chrome. [I've filed a bug with Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=990393&can=2&q=small%20timeConstant). We'll see what they say when someone has a chance to take a look. In the meantime, i think i've got a band-aid. i'll update everyone when i've confirmed it.

adarob:
Thanks Yotam! I know we've been moving slowly on this, but I am hoping we can get a solution to the underlying issue versus doing something hacky-ish in our codebase.

tambien:
@adarob totally agree. waiting for this version to build and if all the tests pass, i'll test out the new version in magenta to see if it fixes the pops

cannoneyed:
@tambien thanks for finding this! It was driving me a bit up the wall trying to figure out where these pops and clicks were coming from.

Anyway, I've opened up another PR addressing the tone.js upgrade, which fixes the glitchy behavior of skipping notes in the beginning of playback by pre-initializing voices in the Player before playback actually begins. Hopefully we can reconcile all of these changes into one fix once Yotam figures out the tone.js pop issue!

https://github.com/tensorflow/magenta-js/pull/327

Thanks all for all the hard work on this!

PR #317, opened 2019-05-31
Throw an error if the visualizer can't size correctly

notwaldorf:
This is mostly whitespace changes + 
- throw an error if we don't know when the sequence ends (we math a lot of NaN sizes in the visualizer in this case)
- stop type checking lib dependencies so that our build doesn't fail because the tfjs types are busted


music/src/core/visualizer.ts
```diff
@@ -135,12 +135,19 @@ export abstract class BaseVisualizer {
 
     // Calculate a nice width based on the length of the sequence we're
     // playing.
+    // Warn if there's no totalTime or quantized steps set, since it leads
+    // to a bad size.
     const endTime = this.sequenceIsQuantized ?
         this.noteSequence.totalQuantizedSteps :
         this.noteSequence.totalTime;
+    if (!endTime) {
+      throw new Error(`The sequence you are using with the \
```
adarob: Is it possible to use + here so the formatting isn't all weird?

PR #318, opened 2019-06-22
Fix compilation error

nuc:
When trying to `yarn build` inside music, we were getting a:

```
Interface 'NumericTensor<R>' incorrectly extends interface 'Tensor<R>'.
  Types of property 'data' are incompatible.
```

According to this [Issue](https://github.com/tensorflow/tfjs/issues/1645), the way to fix it is to add this typescript config.

Cheers

### DIFF
--- a/music/tsconfig.json
+++ b/music/tsconfig.json
@@ -19,7 +19,8 @@
     "pretty": true,
     "noFallthroughCasesInSwitch": true,
     "allowUnreachableCode": false,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "skipLibCheck": true
   },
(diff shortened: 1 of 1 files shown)


nuc:
I signed it!

adarob:
I'm not sure if we want to disable these checks. @notwaldorf can you have a look please?

notwaldorf:
@adarob i have this exact fix in an open PR you’re ignoring :)
My understanding is that the error is in tfjs, and either they fix it or we stop typechecking dependencies (which is probably fine since we can’t fix it?)

adarob:
Well you can approve this without me then :)

PR #325, opened 2019-08-01
Update tone.js to latest version (for compat with web workers, node.js)

cannoneyed:
See https://github.com/tensorflow/magenta-js/issues/309

This PR upgrades `tone.js` to the latest version, in order to facilitate getting the library working in node.js and web workers. As @tambien mentioned, the glitching issue was caused by instantiating the `Polysynth` during playback - this PR switches that instantiation to the constructor of the `Player`... I've verified that all of the players in the demo apps work - However, there might be some issues when using a player with loads of voices, since this fix just routes all of the voices to the one, main default synth. The correct way to do this would be to instantiate a `Player` with a fixed set of synth voices / drum kits as determined by constructor parameters, but this will take a bit more of a refactor... worth it?


notwaldorf:
OMG SO EXCITE. Checking.
Also cc @mattetti since he's been looking at this too.

music/src/core/player.ts
```diff
@@ -400,7 +417,10 @@ export class Player extends BasePlayer {
       bass.volume.value = 5;
       this.synths.set(instrument, bass);
     } else {
-      this.synths.set(instrument, new Tone.PolySynth(10).toMaster());
+      // If no synth is present for the proviced voice, set the voice to the
+      // default synth instrument.
+      const defaultSynth = this.synths.get(this.DEFAULT_SYNTH_INSTRUMENT);
```
mattetti: wouldn't we have an issue here since we could only ever have 2 instruments: bass or the default synth.

let's say I call `getSynth(1)` and then `getSynth(2)` I would get the same default synth in both cases which I don't think is the expected behavior.

cannoneyed: Yeah, as I mentioned in the description this is where things get a bit wonky. The previous behavior was to instantiate a new synth on-demand during playback, which both causes glitches in tone 0.13 and is a bit weird, IMO. The correct way to do this would be to set voices explicitly. I'm happy to move towards doing this, but it's a bigger refactor - figured I'd start with the smallest footprint.

cannoneyed: Also - wow, not sure how I didn't realize there's a whole other massive PR with a lot more context for this same issue. Closing in favor of continuing the conversation there.

PR #326, opened 2019-08-02
upgrading to tone@13.8.21

### LINKED ISSUE #309: update tone.js dependency

tambien:
I was able to fix the clicking issue in previous versions of tone@^13.4.9. This doesn't introduce any API changes. 

Should resolve #309 and address #56.

### DIFF
--- a/music/package.json
+++ b/music/package.json
@@ -13,7 +13,7 @@
     "ndarray-resample": "^1.0.1",
     "protobufjs": "^6.8.6",
     "tonal": "^2.0.0",
-    "tone": "^0.12.80"
+    "tone": "^13.8.21"
   },
   "devDependencies": {
     "@types/clone": "^0.1.30",
--- a/music/src/core/player.ts
+++ b/music/src/core/player.ts
@@ -368,8 +368,18 @@ class DrumKit {
  * A `NoteSequence` player based on Tone.js.
  */
 export class Player extends BasePlayer {
-  private synths = new Map<number, any>();  // tslint:disable-line:no-any
   private drumKit = DrumKit.getInstance();
+
+  private bassSynth = new Tone.Synth({
+    volume: 5,
+    oscillator: { type: 'triangle' }
+  }).toMaster();
+
+  private polySynth = new Tone.PolySynth(10, Tone.Synth, {
+	  envelope : {
+		  attack: 0.001
+	  }
+  }).toMaster();
 
   /**
    * The Tone module being used.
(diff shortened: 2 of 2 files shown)


notwaldorf:
sweet! let me test this real quick!

notwaldorf:
So I listened to a lot of outputs and compared them to https://tensorflow.github.io/magenta-js/music/demos/player.html and friends. I thiiiiink the synth from this patch sounds a bit more attack-y (like comes in louder on every step), but it's on every note and it's predictable, and synths sound alien anyway so i think this might be fine. 

There's also a realistic chance I've gone crazy and imagining this because listening to synth twinkle twinkle 20+ times does that to you.

@adarob can you patch this locally and give it a listen too? and if it's fine we can merge this! 🎉🎉🎉🎉

tambien:
oh yeah you're totally right. i accidentally set the envelope.attack to 0.001 when the default used to be 0.005. I'll remove that line so it'll just remain at the default

cannoneyed:
@adarob @tambien @mattetti 
Just wanted to inject myself real quick in here since it seems like we're all working on the same thing! One thing, @tambien, that may or may not be an issue is that this implementation only creates one PolySynth voice for all instruments in a sequence - this may or may not be an issue (since it's 10 voice polyphonic), and it may not be worth going through all the trouble of pre-instantiating voices like my PR

https://github.com/tensorflow/magenta-js/pull/327

I'd lean towards using this solution, since it's a lot simpler, and having more than 10 voices of the same oscillator playing in a sequence just seems like something that probably doesn't need to be supported anyway.

tambien:
@cannoneyed could also easily make that number larger like 20 if this is an issue.

cannoneyed:
@adarob @notwaldorf  Y'alls call... personally, I'm imagining 20 concurrent oscillators playing every voice in a piece sounding rather.... messy. Capping it at some sensible number like 10 or 16 (because binary) is probably the best tradeoff between what the user expects vs performance

notwaldorf:
could also make that value protected, and you can overriding if you really want to by extending the class

cannoneyed:
Also, can verify that this sounds great. Awesome quick work @tambien!!!

notwaldorf:
yup, that envelope fix makes them sound normal again!

mattetti:
Amazing 😍🙌🎆

notwaldorf:
Thanks!

notwaldorf:
Released this as 1.10.0 🎉🎉🎉

adarob:
Woohoo!!

PR #327, opened 2019-08-02
Update tone.js and refactor player

cannoneyed:
Also see: https://github.com/tensorflow/magenta-js/pull/316

Updates tone.js to the latest version in order to support magenta-js working in Node.js and webworker contexts. Because of some changes in the latest version of tone, we also need to address the way we instantiate / access PolySynth voices in the context of playing a sequence. This PR pre-initializes all voices by adding a protected `initializeVoices` method to the base `BasePlayer` class, which is used by the `Player` subclass to initialize voices corresponding to all instruments in a sequence to be played. 

By moving this initialization out of the `playNote` method, a glitch with playback skipping notes in the beginning of the sequence was fixed. However, per comments by @tambien, Chrome seems to have some issues with the latest tone.js in specific cases when scheduling ramps, which is affecting our `Player` class by introducing small clicks and pops on playback.

Hopefully, we can reconcile all of this after @tambien fixes what he's found!


notwaldorf:
Is this still needed after Yotam's PR?

cannoneyed:
I'd say no, unless we want to support the use case of multiple voices per instrument. Personally, I think this complicates things without making it immediately and obviously more useful for most users. If people want to support additional per-voice instrumentation, I'd argue they ought to roll their own more complex Player subclass.

I'll close!

PR #334, opened 2019-08-20
Add basic tests for MusicVAE

notwaldorf:
Mostly to prove that under ideal conditions this works in node...


notwaldorf:
WDYT about adding this as a test?

notwaldorf:
@adarob just to double check: you've seen that this is downloading an actual model, right? :)

PR #342, opened 2019-09-03
Use a non-negative begin value for tf.slice() in MusicVAE Model.decode().

nsthorat:
TensorFlow.js does not support negative start indexing (nor does TensorFlow python).

I'm going to make a change to tfjs-core to throw a better error message in this case.

Sorry about the weirdness!


notwaldorf:
Thank you so much for figuring this out!! I'll test some samples locally and then merge this.

PR #343, opened 2019-09-04
Write docs to gh-pages branch.

adarob:


music/scripts/generate-docs.sh
```diff
@@ -16,14 +16,24 @@
 
 # Regenerates the docs
 #
-# To run, execute 'yarn doc'.
+# To run, execute 'yarn docs'.
+
+# Exit on error.
+set -e
 
 # Generate the docs.
```
notwaldorf: technically this comment should go above the npx line, these first lines are just setting up variables

music/scripts/generate-docs.sh
```diff
@@ -49,3 +59,16 @@ for path in ./*.html; do
     sed -i "" "s%${search}%${replace}%g" $path
   fi
 done
+
+cd $currDir
+yarn build-demos
+mkdir -p $TMP_DIR/demos && cp demos/*.{js,html,mid,css} $TMP_DIR/demos
+git checkout gh-pages
```
notwaldorf: Maybe a quick summary of what this is doing?

music/scripts/generate-docs.sh
```diff
@@ -49,3 +59,16 @@ for path in ./*.html; do
...
+docsDir=$(git rev-parse --show-toplevel)
+cd $baseDir
+git rm -fr music
+rsync -a $TMP_DIR/ music/
```
notwaldorf: why rsync and not something like cp?

music/scripts/generate-docs.sh
```diff
@@ -49,3 +59,16 @@ for path in ./*.html; do
...
+git add music
+git commit -m "Updating music docs: $currDate"
```
notwaldorf: i would create that currDate variable here, not up top (i forgot what it did already)

music/scripts/generate-docs.sh
```diff
@@ -16,14 +16,24 @@
...
-# To run, execute 'yarn doc'.
+# To run, execute 'yarn docs'.
+
+# Exit on error.
+set -e
 
 # Generate the docs.
-npx typedoc --tsconfig "tsconfig.es5.json" --sourcefile-url-prefix 'https://github.com/tensorflow/magenta-js/tree/master/music/src/' --out ../docs/music src --mode modules --excludePrivate --exclude '**/*+(index|test|lib).ts' --excludeExternals
+TMP_DIR=/tmp/mm_docs
+currRepo=$(git rev-parse --abbrev-ref HEAD)
```
notwaldorf: i think you mean currBranch not repo, right?

music/scripts/generate-docs.sh
adarob: So that we don't clobber untracked files from the other branch.

PR #347, opened 2019-09-05
Use shared doc generation script.

adarob:


scripts/generate-docs.sh
```diff
@@ -21,16 +21,27 @@
 # Exit on error.
 set -e
 
+PKG_NAME=$1
```
notwaldorf: Please document what the new arguments mean now (and how you're supposed to call this script)

image/package.json
```diff
@@ -27,8 +27,7 @@
         "lint": "tslint -c ../tslint.json -p . -t verbose",
         "build-demos": "cd demos && ./build-demos.sh",
         "run-demos": "yarn build-demos && http-server demos/",
-        "publish-demos": "yarn build-demos && mkdir -p ../docs/image/demos && cp demos/*.{js,html,css,jpg} ../docs/image/demos",
-        "doc": "typedoc --sourcefile-url-prefix 'https://github.com/tensorflow/magenta-js/tree/master/image/src/arbitrary_stylization/' --out ../docs/image src --mode file --excludePrivate --exclude '**/*+(index|test).ts' --excludeExternals && yarn publish-demos"
+        "docs": "sh ../scripts/generate-docs.sh 'image' 'https://github.com/tensorflow/magenta-js/tree/master/image/src/arbitrary_stylization/' 'file'"
```
notwaldorf: i'm not 100% convinced why this file couldn't also be "modules", btw. have you tried? then you wouldn't have to pass this in

scripts/generate-docs.sh
```diff
@@ -21,16 +21,27 @@
...
+URL_PREFIX=$2
+MODE=$3
```
notwaldorf: i think you should default `MODE` to `modules`, like you default the tsconfig file

scripts/generate-docs.sh
notwaldorf: also i thiiiiiink you could get the package name from the url prefix btw (it's the one inside `magenta-js/tree/master/($1)/src/`, but your call

PR #350, opened 2019-09-09
Document using your own soundfont

### LINKED ISSUE #297: How can I use local soundfonts?
Hello, I would like to have other instruments for the SoundFontPlayer class, how can I do that? Thank you.

notwaldorf:
Fixes #297.

Let me know if some of this doesn't make sense, or if you have some extra details to add (since you made most of our SoundFonts, I think)

### DIFF
--- a/music/src/core/soundfont.ts
+++ b/music/src/core/soundfont.ts
@@ -347,10 +347,53 @@ export interface SoundFontSpec {
   instruments: InstrumentsSpec;
 }
 
+// tslint:disable:max-line-length
 /**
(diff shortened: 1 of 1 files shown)


music/src/core/soundfont.ts
```diff
@@ -347,10 +347,53 @@ export interface SoundFontSpec {
   instruments: InstrumentsSpec;
 }
 
+// tslint:disable:max-line-length
```
adarob: please do this just for next lines

notwaldorf: I do...I re-enable it on L395

PR #358, opened 2019-09-19
add a demo page for what the soundfonts sound like

notwaldorf:
Also added some licenses to some demos that didn't have licenses 😲


notwaldorf:
actually hold off on reviewing this for a second, turns out i had already added a section for this on a different page.

notwaldorf:
@adarob ok this is ready for review for realsies

PR #362, opened 2019-09-25
Docs for 1.11.2

notwaldorf:


notwaldorf:
Why are there so many changes? @adarob

PR #365, opened 2019-10-01
fix commonjs mappings for shared deps

developit:
/cc @notwaldorf


notwaldorf:
Yesssss thank you Jason!!! You are too good for this world.

PR #367, opened 2019-10-03
Replace console.logs with logging logs

naclcaleb:
Not sure if I put these in all the write places. 

I didn't know what to do with logs with a second parameter (i.e., `console.log("Something went wrong", err)`), so I just left them.

### DIFF
--- a/music/src/core/recorder.ts
+++ b/music/src/core/recorder.ts
@@ -24,6 +24,8 @@ import {NoteSequence} from '../protobuf/index';
 
 import {DEFAULT_QUARTERS_PER_MINUTE} from './constants';
 
+import * as logging from './logging';
+
 /**
  * An interface for providing configurable properties to a Recorder.
  * @param qpm The tempo at which to play the click track.
@@ -135,7 +137,7 @@ export class Recorder {
   }
 
   private midiReady(midi: WebMidi.MIDIAccess) {
-    console.log('Initialized Recorder');
+    logging.log('Initialized Recorder');
     const inputs = midi.inputs.values();
     for (let input = inputs.next(); input && !input.done;
          input = inputs.next()) {
--- a/music/src/core/recorder.ts
+++ b/music/src/core/recorder.ts
@@ -137,7 +137,7 @@ export class Recorder {
   }
 
   private midiReady(midi: WebMidi.MIDIAccess) {
-    logging.log('Initialized Recorder');
+    logging.log('Initialized Recorder', 'Recorder');
     const inputs = midi.inputs.values();
     for (let input = inputs.next(); input && !input.done;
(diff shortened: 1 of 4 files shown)


naclcaleb:
Addresses #367

music/src/core/recorder.ts
```diff
@@ -135,7 +137,7 @@ export class Recorder {
   }
 
   private midiReady(midi: WebMidi.MIDIAccess) {
-    console.log('Initialized Recorder');
+    logging.log('Initialized Recorder');
```
notwaldorf: I would use "Recorder" as the prefix here (and everywhere where the prefix could be helpful)

music/src/core/soundfont.ts
```diff
@@ -194,7 +195,7 @@ export class Instrument {
       sampleNamesAndURLs.forEach(
           (nameAndURL) => this.buffers.add(nameAndURL.name, nameAndURL.url));
       await new Promise(resolve => Tone.Buffer.on('load', resolve));
-      console.log(`Loaded samples for ${this.name}.`);
+      logging.log(`Loaded samples for ${this.name}.`);
```
notwaldorf: I'd probably use SoundFont here as the prefix

naclcaleb: Should I use that prefix for all the logs in the file?

notwaldorf: yeah, i think that would be neat (it helps see where the logs are coming from)

naclcaleb: Ok, I added a new commit.

Should I add this to the other files as well?

notwaldorf: yes please!

naclcaleb: Ok, does my latest commit look ok?

naclcaleb:
Ok, I finally got it to pass the checks. Is this satisfactory?

music/src/core/soundfont.ts
```diff
@@ -169,7 +170,7 @@ export class Instrument {
         samples
             .filter((info) => {
               if (info.pitch < this.minPitch || info.pitch > this.maxPitch) {
-                console.log(
+                logging.log(
                     `Pitch ${info.pitch} is outside the valid range for ${
                         this.name}, ignoring.`);
```
notwaldorf:
```suggestion
                        this.name}, ignoring.`, 'SoundFont');
```

naclcaleb:
Ah, whoops! Thanks for the suggestion; I committed it. But now there's an error about Google CLA...

notwaldorf:
No worries, it's just mad at me not you :)

notwaldorf:
@googlebot I fixed it.

PR #392, opened 2020-01-08
Fixed internal link on paragraph 51- Duplicate word on line 56

kant:
* semantic issue on line 77


notwaldorf:
Uhhhhhh there's now a legit MusicRNN failure apparently? This makes no sense :/

notwaldorf:
Im going to hope the failure is a flake. Even if it isn't, there's no way this PR is breaking that so i'll take a look separately.

PR #399, opened 2020-02-05
Note data attributes in SVG piano roll

cifkao:
Enhancing the `rect` elements in `PianoRollSVGVisualizer` with data attributes like `data-program`, `data-is-drum`, `data-instrument`.

Now we can style notes by instrument using CSS and have colorful piano rolls like this:
image

### DIFF
--- a/music/src/core/visualizer.ts
+++ b/music/src/core/visualizer.ts
@@ -324,10 +324,7 @@ export class Visualizer extends PianoRollCanvasVisualizer {
 /**
  * HTML data attribute key-value pair.
  */
-interface DataAttribute {
-  key: string;
-  value: any; // tslint:disable-line:no-any
-}
+type DataAttribute = [string, any]; // tslint:disable-line:no-any
 
 /**
  * Displays a pianoroll as an SVG. Pitches are the vertical axis and time is
@@ -430,12 +427,12 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
       const note = this.noteSequence.notes[i];
       const size = this.getNotePosition(note, i);
       const fill = this.getNoteFillColor(note, false);
-      const dataAttributes = [
-        {key: 'index', value: i},
-        {key: 'instrument', value: note.instrument},
-        {key: 'program', value: note.program},
-        {key: 'isDrum', value: note.isDrum === true},
-        {key: 'pitch', value: note.pitch},
+      const dataAttributes : DataAttribute[] = [
+        ['index', i],
+        ['instrument', note.instrument],
+        ['program', note.program],
+        ['isDrum', note.isDrum === true],
+        ['pitch', note.pitch],
       ];
 
       this.drawNote(size.x, size.y, size.w, size.h, fill, dataAttributes);
@@ -467,9 +464,9 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
     rect.setAttribute('y', `${Math.round(y)}`);
     rect.setAttribute('width', `${Math.round(w)}`);
     rect.setAttribute('height', `${Math.round(h)}`);
-    dataAttributes.forEach((attr: DataAttribute) => {
-      if (attr.value !== undefined) {
-        rect.dataset[attr.key] = `${attr.value}`;
+    dataAttributes.forEach(([key, value]: DataAttribute) => {
+      if (value !== undefined) {
+        rect.dataset[key] = `${value}`;
       }
     });
     this.svg.appendChild(rect);
--- a/music/src/core/visualizer.ts
+++ b/music/src/core/visualizer.ts
@@ -320,6 +320,15 @@ export class Visualizer extends PianoRollCanvasVisualizer {
         'mm.Visualizer', logging.Level.WARN);
   }
 }
+
+/**
+ * HTML data attribute key-value pair.
+ */
+interface DataAttribute {
+  key: string;
+  value: any; // tslint:disable-line:no-any
+}
+
 /**
  * Displays a pianoroll as an SVG. Pitches are the vertical axis and time is
  * the horizontal. When connected to a player, the visualizer can also highlight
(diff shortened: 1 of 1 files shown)


music/src/core/visualizer.ts
```diff
@@ -437,7 +444,8 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
   }
 
   private drawNote(
-      x: number, y: number, w: number, h: number, fill: string, index: number) {
+      x: number, y: number, w: number, h: number, fill: string,
+      dataAttributes: {[attr: string]: any}) { // tslint:disable-line:no-any
```
notwaldorf: wahhhh I know this is annoying, but could you please make a type for this (maybe `SVGNoteDataAttributes`?). You can look at the definition of `VisualizerConfig` for what that looks like.

cifkao: So like this?
```typescript
interface SVGNoteDataAttributes {
  index: number,
  instrument: number,
  ...
}
```
But can we then iterate over the attributes [here](https://github.com/tensorflow/magenta-js/pull/399/files/c343792292d8701c19bb6605af30d738c728db92#diff-cde90f4f7fada5c5c2fb344599c23cbeR461)? If not, then I think it would be better to just pass the actual note in.

Also, it might better to use `rect.dataset.isDrum` instead of `rect.setAttribute('is-drum')` ([`dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset) takes care of translating the camel-cased names to the dash-separated ones).

notwaldorf: Oh no, i meant more like:
(...fill: string, dataAttributes: []: SVGNoteAttribute)
interface SVGNoteDataAttributes {
  key: String,
  value: String
}

And then you can iterate over the array.

And yeah, definitely use dataset. my bad, didn't see that the first time around.

cifkao: OK, got it. Am I allowed to do `value: any` (suppressing `no-any`)? Otherwise I guess I have to `.toString()` everything (and first check if it's defined).

Sorry, this is my first time using TypeScript. :sweat_smile:

cifkao: I mean, I would have to do something like this for every attribute:
```typescript
if (note.instrument !== undefined) {
  dataAttributes.push({key: 'instrument', value: `${note.instrument}`});
} 
```
(Because in the visualizer demo, it does happen that `note.instrument === undefined`.)

notwaldorf: you can push them as undefined and then not set them, no? you already check if they're undefined before setting them

notwaldorf: Typescript enforces you to cast them to string before you put them in the dataset anyway (so `foo.dataset.batman = String(someNumber)`, so I don't think you need to do the type `any` in there

cifkao: Yes, but this way `value` cannot be `string`, because I'm passing in numbers and booleans (and `undefined`s)...

cifkao: Do you want me to push it like that so you can have a look?

cifkao: If I turn it into a string before passing it into `drawNote`, I need to check for `undefined`, otherwise it will get turned into `"undefined"`, won't it? And if want to do it inside `drawNote`, I need the `any` to pass it in.

cifkao: And I seem to be able to do `rect.dataset[key] = value` without TypeScript complaining. :thinking:

music/src/core/visualizer.ts
```diff
@@ -437,7 +453,8 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
   }
 
   private drawNote(
-      x: number, y: number, w: number, h: number, fill: string, index: number) {
+      x: number, y: number, w: number, h: number, fill: string,
+      dataAttributes: DataAttribute[]) {
```
notwaldorf: actually, i just double checked and I think you could do [{x: number, y: any}] btw, which seems a bit better? Sorrrrrrrry :(

cifkao: Do you mean something like this? I was not able to find what the curly and square bracket thing meant.

music/src/core/visualizer.ts
```diff
@@ -450,7 +464,11 @@ export class PianoRollSVGVisualizer extends BaseVisualizer {
     rect.setAttribute('y', `${Math.round(y)}`);
     rect.setAttribute('width', `${Math.round(w)}`);
     rect.setAttribute('height', `${Math.round(h)}`);
-    rect.setAttribute('data-index', `${index}`);
+    dataAttributes.forEach(([key, value]: DataAttribute) => {
+      if (value !== undefined) {
+        rect.dataset[key] = `${value}`;
```
notwaldorf: this works even though `value` is a number because you are explicitely converting it to a string here. This is why I don't think you really need that `any` ...

music/src/core/visualizer.ts
notwaldorf: I thought [{x:number, y:number}] would say that it's an array of objects of that form, but maybe im wrong!

music/src/core/visualizer.ts
cifkao: Maybe I'm missing something, but I still don't know what you want me to do instead. If I just do `type DataAttribute = [string, string]`, then TypeScript will make me convert to string in `draw()` rather than here. This also means I need to check for `undefined` in `draw` *as well as here*, i.e.:
```typescript
      const dataAttributes : DataAttribute[] = [
        ['index', i],
        ['instrument', note.instrument === undefined ? undefined : `${note.instrument}`],
        ...
      ];
```

In that case it would be easier to just `return rect` from here and then in `draw()`, do
```typescript
if (note.instrument !== undefined) rect.dataset.instrument = `${note.instrument}`;
...
```

Or, if you're OK with setting `data-instrument="undefined"`, then I don't need to check for it at all.

notwaldorf: Let's just leave it how it is, I think it's fine enough.

PR #404, opened 2020-02-17
Prevent playback after stop()

### LINKED ISSUE #380: `run` callback called after `Player.stop()`
`SoundFontPlayer` sometimes continues the playback for a very short time (invoking the `run` callback) after the `stop` method has been called. This typically happens when there is a rapid sequence of notes.

This can cause problems for example when using the player together with a `Visualizer` and trying to `clear` it or display something else on the same canvas immediately after calling `stop`. In this case, the old `NoteSequence` will overwrite the new content of the canvas.

cifkao:
Should fix #380.

### DIFF
--- a/music/src/core/player.ts
+++ b/music/src/core/player.ts
@@ -161,7 +161,7 @@ export abstract class BasePlayer {
     }
 
     const thisPart = new Tone.Part((t: number, n: NoteSequence.INote) => {
-      // Prevent playback after the part has been removed
+      // Prevent playback after the part has been removed.
       if (this.currentPart !== thisPart) {
         return;
       }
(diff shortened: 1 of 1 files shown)


music/src/core/player.ts
```diff
@@ -160,7 +160,12 @@ export abstract class BasePlayer {
       throw new Error('Cannot specify a `qpm` for a non-quantized sequence.');
     }
 
-    this.currentPart = new Tone.Part((t: number, n: NoteSequence.INote) => {
+    const thisPart = new Tone.Part((t: number, n: NoteSequence.INote) => {
+      // Prevent playback after the part has been removed
```
notwaldorf:
```suggestion
      // Prevent playback after the part has been removed.
```

Issue #393, opened 2020-01-14
Export STFT?


deltonvaz:
Hello! 

I have a project and I need to use the [Short Term Fourier Transform](https://tensorflow.github.io/magenta-js/music/modules/_core_audio_utils_.html#stft) directly but I noticed that despite being the documentation I cannot use it. Would it be possible to export the function so that it could be used? Or is there another way to access it that I'm not seeing?

Thanks õ/

notwaldorf:
Oh yeah, I think we can do that! Going to send a PR out today

notwaldorf:
Alright, I've thought about this more and I don't think we want to add all these audio_utils as a public API -- we added them for a model in particular, and the problem is that they aren't tested for all cases. We definitely don't plan on working on them or adding new features, and there's many libraries out there that give you different ways to do stfts. So this is why I think they should remain a private API.

I'm not exactly sure why the documentation is being generated for them (I assume it's a bug with the docs generator), but I think your best bet if you want to use this particular stft implementation is to just copy it into your own project from [here](https://github.com/tensorflow/magenta-js/blob/master/music/src/core/audio_utils.ts#L205)

PR #405, opened 2020-02-20
export the audio utils

notwaldorf:
Fixes https://github.com/tensorflow/magenta-js/issues/393

(I don't remember if there's a reason why we don't explicitly export the audio_utils btw)


adarob:
Makes me a bit nervous since these aren't tested though...

notwaldorf:
What if I _just_ export `stft()`, which is what the issue asks for?

PR #409, opened 2020-02-21
Fix docs script

notwaldorf:
Fixed a slew of things in our docs. Thank you for coming to my ted talk, grab a coffee cause this is a doozy.

## Necessary changes to make our docs work again
- at some point `typedoc-plugin-sourcefile-url` stopped working for us and after 2 hours of trying to fix it I just replaced it with a `sed`

## Inoffensive changes because I spent so much time staring at the typedoc docs
- added `includeVersion` which now prints the package version in the docs. It's kind of nice to keep track of what you're looking at.
- changed it to generate docs for exported functions (vs every single function in a class). I hoped this would stop generating docs for `audio_utils`. It didn't, but it did reduce the number of uncallable functions we document in `audio_utils` (sigh). 
- added typedoc files to each project to simplify the script (so that the script is all hacks now)
- made git stop dumping every file it removes and checks out to the console 
- changed it to use the es6 tsconfig which made no difference

## Improve docs!
- We were showing docs for a bunch of things we weren't actually exporting (like `coconet_utils`, or `audio_utils`, etc), so I added a mondo hack for just the `music` package to scrape the `index.ts` file and only show the exported modules in the left hand table of contents. I think this is what we want, and I am 100% sure typedoc doesn't do it out of the box.

Here's what the docs look like now: https://meowni.ca/magenta-js/music/index.html

In particular, before this PR
Screen Shot 2020-02-20 at 11 00 57 PM

after this PR
Screen Shot 2020-02-20 at 11 01 18 PM


notwaldorf:
I think this is strictly better than what we have now because this actually shows NoteSequences: https://meowni.ca/magenta-js/music/modules/_protobuf_.html, whereas https://tensorflow.github.io/magenta-js/music/modules/_protobuf_.html doesn't

scripts/generate-docs.sh
```diff
@@ -33,31 +33,67 @@ currDir=$(pwd)
...
+</script>"
 fi
 
 # Generate the docs.
 rm -rf $tmpDir
-npx typedoc --tsconfig $tsconfig --sourcefile-url-prefix $urlPrefix --out $tmpDir  --mode $mode --excludePrivate --exclude '**/*+(index|test|lib).ts' --excludeExternals src
+npx typedoc --options typedoc.json src --out $tmpDir
+
+###
+# Here begin the different hacks to fix the docs because typedoc has a lot of bugs.
```
Gerrit0: Fair but ouch :/ its what happens when a project gets next to no attention for a couple years.

scripts/generate-docs.sh
```diff
@@ -33,31 +33,67 @@ currDir=$(pwd)
...
+###
+
+# Typedoc generates documentation for _all_ files, not just the ones
+# actually exported in the library, so insert a script to fix the index.html.
+# see https://github.com/TypeStrong/typedoc/issues/639
+echo $scriptToFixTheToc >> $tmpDir/index.html
```
Gerrit0: The builtin TOC option requires that the names be exact, including the quotes since they are module names.

What you are after is this:

```
{
  "toc": [
    "\"coconet\"",
    "\"core\"",
    ....
  ],
  other options
}
```

scripts/generate-docs.sh
notwaldorf: Oh, no worries! I know documentation projects are often understaffed :(

scripts/generate-docs.sh
notwaldorf: Oh brilliant, would you mind if I sent a PR to update this in the docs? This is super confusing

Gerrit0: That would be great!

adarob:
Thanks for doing this!!!

PR #416, opened 2020-02-24
Remove check for contiguous instrument numbers in midi_io

cifkao:
I ran into this error ("Instrument list must be continuous and start at 0") when I was trying to save a `NoteSequence` from which I had removed some notes, which created a gap in the instrument numbers.

I don't see the purpose of this check, so I'm proposing to get rid of it. Looking at the history, it seems to have been there because of the way channel numbers were assigned. That is now different, so it should be safe to remove it, shouldn't it?


cifkao:
This is unnecessary once #422 is merged.

notwaldorf:
I merged #422, so closing this

PR #418, opened 2020-02-27
Correct behavior with multiple players

### LINKED ISSUE #403: Multiple players getting mixed up
Multiple players on one page don't seem to play well with each other. For example, in the `callback_player` demo, try clicking "play twinkle with click" first and while it is playing, click "play teapot without click". The following happens:

- When the first player stops, the other one stops as well. However, only the callback for the first player gets called.
- If you click "play twinkle with click" again, both sequences will play simultaneously in the same way as the first time (unless you click the stop button for the other sequence first).

Is it possible to implement players in such a way that they can play simultaneously? Otherwise I think possible solutions are:

- Whenever a player wants to start playing, it should first stop any other player (and call its `stop` callback).
- Prevent the player from starting if another player is playing.
- Advise users to not use more than one player at a time.

cifkao:
This PR adds checks to `BasePlayer` to make sure only one player is active at a given time. An error is thrown if an action would put the player in an inconsistent state. Fixes #403.

### DIFF
--- a/music/src/core/player.ts
+++ b/music/src/core/player.ts
@@ -139,9 +139,17 @@ export abstract class BasePlayer {
    * specified, will use either the qpm specified in the sequence or the
    * default of 120. Only valid for quantized sequences.
    * @returns a Promise that resolves when playback is complete.
+   * @throws {Error} If this or a different player is currently playing.
    */
 
   start(seq: INoteSequence, qpm?: number): Promise<void> {
+    if (this.isPlaying()) {
+      throw new Error(`Cannot start playback while "${this.getPlayState()}".`);
+    }
+    if (Tone.Transport.state !== 'stopped') {
+      throw new Error(`Cannot start playback while \`Tone.Transport\` is in use.`);
+    }
+
     this.resumeContext();
     const isQuantized = sequences.isQuantizedSequence(seq);
     if (this.playClick && isQuantized) {
@@ -201,7 +209,7 @@ export abstract class BasePlayer {
    * Stop playing the currently playing sequence right away.
    */
   stop() {
-    if (this.currentPart) {
+    if (this.isPlaying()) {
       this.currentPart.stop();
       Tone.Transport.stop();
       this.currentPart = null;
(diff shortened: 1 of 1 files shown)


music/src/core/player.ts
```diff
@@ -139,9 +139,18 @@ export abstract class BasePlayer {
    * specified, will use either the qpm specified in the sequence or the
    * default of 120. Only valid for quantized sequences.
    * @returns a Promise that resolves when playback is complete.
+   * @throws {Error} If this or a different player is currently playing.
    */
 
   start(seq: INoteSequence, qpm?: number): Promise<void> {
+    if (this.isPlaying()) {
+      throw new Error(`Cannot start playback while "${this.getPlayState()}".`);
```
notwaldorf: I think you can just say "Cannot start playback; player is already playing" to be more clear.

music/src/core/player.ts
```diff
@@ -212,25 +221,36 @@ export abstract class BasePlayer {
...
+    if (!this.isPlaying()) {
+      throw new Error('Cannot pause playback while the player is stopped.');
+    }
     Tone.Transport.pause();
   }
 
   /**
-   * Pause playing the currently playing sequence right away. Call resume()
-   * to resume playing the sequence.
+   * Resume playing the sequence after pause().
```
notwaldorf: Wow, thanks for fixing these comments. I don't know how I shipped them like this 🤦‍♀

music/src/core/player.ts
```diff
@@ -201,7 +210,7 @@ export abstract class BasePlayer {
    * Stop playing the currently playing sequence right away.
    */
   stop() {
-    if (this.currentPart) {
+    if (this.isPlaying()) {
```
notwaldorf: I don't think this is correct. The player could be paused, and you should still be allowed to stop it.

music/src/core/player.ts
```diff
@@ -212,25 +221,36 @@ export abstract class BasePlayer {
...
+    }
     Tone.Transport.start();
   }
 
   /**
    * Seek to a number of seconds in the NoteSequence.
+   * @throws {Error} If the player is stopped.
    */
   seekTo(seconds: number) {
+    if (!this.isPlaying()) {
```
notwaldorf: Not sure about this one either. You can definitely seek while the player is paused or stopped. I would remove this

music/src/core/player.ts
cifkao: `!!this.currentPart` is the definition of `isPlaying()`, so this behaves the same way as before. I just thought it was clearer to write it like this.

music/src/core/player.ts
cifkao: Same as above: `isPlaying()` returns `true` when the player is paused. And when it's stopped, you can't `resume()` it anymore; you can only call `start()` which will start playing from time 0. Am I correct?

cifkao: In any case, we don't want to allow doing anything with the `Transport` when the player is stopped, because it might affect some other player.

music/src/core/player.ts
notwaldorf: ah, good call.

notwaldorf:
(I'll do a release now that has all of your recent changes, thanks so much for them!)

PR #421, opened 2020-02-29
Make sure all notes are at least 1px wide

cifkao:
Adding a `minNoteLength` parameter (1 px by default) to make sure very short notes are visible and don't cause errors when they are shorter than `noteSpacing`.


notwaldorf:
In your use of this do you use a non-1px value? I would be very ok with making sure the width is always at least 1, and not exposing it as a new config option

cifkao:
Sure, 1px is fine.

PR #428, opened 2020-03-11
Fix sequences.trim

cifkao:
`trim` was using the unshifted `end` time for note truncation and for setting `totalTime`.

### DIFF
--- a/music/src/core/sequences.ts
+++ b/music/src/core/sequences.ts
@@ -656,15 +656,16 @@ function trimHelper(
           (truncateEndNotes || n[endKey] <= end));
 
   // Shift the sequence to start at 0.
+  result[totalKey] -= start;
   for (let i = 0; i < result.notes.length; i++) {
     result.notes[i][startKey] -= start;
     result.notes[i][endKey] -= start;
 
     if (truncateEndNotes) {
-      result.notes[i][endKey] = Math.min(result.notes[i][endKey], end);
+      result.notes[i][endKey] = Math.min(result.notes[i][endKey],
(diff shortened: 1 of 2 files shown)


cifkao:
Do you think you could have a look at this, @notwaldorf? This bug causes trimmed note sequences to have the wrong `totalTime` if `start > 0`.

music/src/core/sequences.ts
```diff
@@ -656,15 +656,16 @@ function trimHelper(
...
 
   // Shift the sequence to start at 0.
+  result[totalKey] -= start;
   for (let i = 0; i < result.notes.length; i++) {
     result.notes[i][startKey] -= start;
     result.notes[i][endKey] -= start;
 
     if (truncateEndNotes) {
-      result.notes[i][endKey] = Math.min(result.notes[i][endKey], end);
+      result.notes[i][endKey] = Math.min(result.notes[i][endKey],
```
notwaldorf, line 665: Yeah, i think you're right, and this makes it consistent with https://github.com/tensorflow/magenta/blob/master/magenta/music/sequences_lib.py#L124

Do you need to update .totalTime too? as per https://github.com/tensorflow/magenta/blob/master/magenta/music/sequences_lib.py#L126

cifkao: Should be already updated above (lines 653 and then 659). (I'm shifting it before, so that I can use it on the line below.)

adarob:
@notwaldorf is this ready to merge?

notwaldorf:
Yup! My bad.

PR #431, opened 2020-03-29
Updated @magenta/music dependencies

Seanitzel:
Updated all dependencies.


Seanitzel:
@googlebot I signed it!

notwaldorf:
Thanks for the PR! Unfortunately, I'm not super sure we need it. The ^ basically means "give me any minor/patch version that matches the major version of this" so all of these deps would get installed anyway, even theyre not listed explicitely in the package.json. I think there's just too many deps that update all the time, and enforcing this will always be up to date isn't useful or feasible.

PR #440, opened 2020-04-22
add Melody class for quantized monophonic melodies

iansimon:


music/src/core/melodies.ts
```diff
@@ -0,0 +1,142 @@
...
+import {INoteSequence, NoteSequence} from '../protobuf/index';
+
+import * as sequences from './sequences';
+
+export const NO_EVENT = 0;
+export const NOTE_OFF = 1;
+const FIRST_PITCH = 2;
+
+/**
+ * Melody representation with note onsets, `NO_EVENT` (sustain) and `NOTE_OFF`.
```
notwaldorf, line 32: how is a note onset represented?

notwaldorf: and is this different than the Performance representation?

music/src/core/melodies.ts
```diff
@@ -0,0 +1,142 @@
...
+          if (n1.quantizedStartStep === n2.quantizedStartStep) {
+            return n2.pitch - n1.pitch;
+          }
+          return n1.quantizedStartStep - n2.quantizedStartStep;
+        });
+    const events = new Int32Array(numSteps || noteSequence.totalQuantizedSteps);
+    let lastStart = -1;
+    sortedNotes.forEach(n => {
+      if (n.quantizedStartStep === lastStart) {
+        if (!ignorePolyphony) {
```
notwaldorf, line 80: I would add this as a helper in sequences, like `assertIsMonophonic`

notwaldorf: (just how you have `assertIsQuantized` above

music/src/core/sequences.ts
```diff
@@ -402,6 +402,19 @@ export function unquantizeSequence(qns: INoteSequence, qpm?: number) {
   return ns;
 }
 
+/**
+ * Create an empty quantized NoteSequence with steps per quarter note and tempo.
+ * @param stepsPerQuarter The number of steps per quarter note to use.
+ * @param qpm The tempo to use.
+ * @returns A new quantized NoteSequence.
+ */
+export function createQuantizedNoteSequence(
```
notwaldorf, line 411: this doesn't feel like a helper here (it's too generic), maybe just in the class you're using it in?

music/src/core/melodies.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Melody representation with note onsets, `NO_EVENT` (sustain) and `NOTE_OFF`.
+ *
+ * @param events An array of melody events.
+ * @param minPitch The minimum pitch to represent.
+ * @param maxPitch The maximum pitch to represent.
+ *
+ */
+export class Melody {
+  readonly events: Int32Array;
```
notwaldorf, line 40: is that the same as `Number[]` ?

music/src/core/melodies.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Extract a melody from a `NoteSequence`.
+   *
+   * @param noteSequence `NoteSequence` from which to extract a melody.
+   * @param minPitch The minimum pitch to represent. Those above this value will
+   * cause an error to be thrown.
+   * @param maxPitch The maximum pitch to represent. Those above this value will
+   * cause an error to be thrown.
+   * @param ignorePolpyhony If false, an error will be raised when notes start
```
notwaldorf, line 58: typo: polyphony

music/src/core/melodies.ts
notwaldorf: The fact that I'm confused leads me to believe "melody" is too generic a name

music/src/core/melodies.ts
iansimon: One issue is that isQuantized is a constant time check whereas isMonophonic requires checking the start times of all the notes, so here (I just moved this code from data.ts) it's done in the already-existing melody conversion loop.

music/src/core/melodies.ts
iansimon: I think it is not even though I have forgotten the small amount of TypeScript I knew.  I tried using a number[] but the tf.tensor function didn't want to take one of those (in data.py) so to avoid multiple array conversions I just used Int32Array.  Is that weird?

music/src/core/sequences.ts
iansimon: But now it's being used in two places: melodies.ts and data.ts.

music/src/core/melodies.ts
notwaldorf: ohhh i see, yeah got it.

music/src/core/sequences.ts
notwaldorf: 👍

music/src/core/melodies.ts
iansimon: Added some doc to help clarify.

iansimon:
@cjcai I cannot add you as an official reviewer but you can take a look if you like.

music/src/core/melodies.ts
iansimon: Fixed.

iansimon:
> (but your tests are on fire)

They are now not on fire!

PR #441, opened 2020-04-23
add rhythm and shape melody control signals

iansimon:


iansimon:
@cjcai feel free to take a look if you like.

music/src/core/melodies.ts
```diff
@@ -144,3 +145,112 @@ export class Melody {
...
+export interface MelodyControl {
+  readonly depth: number;
+  extract(melody: Melody): tf.Tensor2D;
+}
+
+/**
+ * Rhythm control signal. Extracts a depth-1 tensor indicating note onsets.
+ */
+export class MelodyRhythm implements MelodyControl {
+  readonly depth = 1;
```
notwaldorf, line 165: i think this can just be private then? does anyone need to actually read it?

music/src/core/melodies_test.ts
```diff
@@ -58,3 +59,24 @@ test('Test Melody To NoteSequence', (t: test.Test) => {
   t.deepEqual(ns, expected);
   t.end();
 });
+
+test('Test Melody Rhythm Control', (t: test.Test) => {
+  const events = new Int32Array([0, 2, 0, 1, 4, 6, 9, 1]);
+  const melody = new Melody(events, 60, 72);
```
notwaldorf, line 65: i feel this is a bit weird, and you should just be able to pass a normal Javascript array [ ] to that Melody constructor. Are the types getting in the way because it needs to not contain floats?

iansimon: I switched the Melody constructor to take ArrayLike<number>; is that more normal?  I don't want to have it just take number[] because tfjs gives back data as Int32Array.

music/src/core/melodies.ts
iansimon: It's the dimension of the returned arrays, which really feels to me like something that should be public.

PR #447, opened 2020-05-04
support extra control signals in MusicVAE and add control preprocessing bidirectional LSTM

iansimon:


iansimon:
Also may be of interest to @cjcai

notwaldorf:
I'm actually gonna pass this baton to @adarob since MusicVAE is his baby

adarob:
baton accepted!

music/src/music_vae/model.ts
```diff
@@ -71,11 +71,65 @@ abstract class Encoder {
...
+    const forgetBias = tf.scalar(1.0);
+    const lstm = (data: tf.Tensor2D, state: [tf.Tensor2D, tf.Tensor2D]) =>
+        tf.basicLSTMCell(
+            forgetBias, lstmVars.kernel, lstmVars.bias, data, state[0],
+            state[1]);
+    const splitInputs = tf.split(inputs.toFloat(), length, 1);
+    const outputStates: tf.Tensor2D[] = [];
+    for (const data of (fw ? splitInputs : splitInputs.reverse())) {
+      state = lstm(data.squeeze([1]) as tf.Tensor2D, state);
+      outputStates.push(state[1]);
```
adarob, line 122: maybe we should call this outputStatesH since we're not keeping the C?

adarob: Actually we can also just document this better ... like above we can say "A batch of forward and backward output (h) LSTM states."

music/demos/index.html
```diff
@@ -42,6 +42,7 @@ <h3>
         <li><a href="music_vae.html">Generate new NoteSequences using MusicVAE</a></li>
         <li><a href="midi_me.html">Personalize your own MusicVAE model with MidiMe</a></li>
         <li><a href="multitrack.html">Multitrack MusicVAE</a></li>
+        <li><a href="melody_controls.html">Melody MusicVAE with conditioning</a></li>
```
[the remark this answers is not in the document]
iansimon: The only really obvious subgroup was MusicVAE so I added a sublist for that one.

music/src/music_vae/model.ts
iansimon: My LSTM recollection is rusty but "h" is the same as "output", right?  And "c" is the internal state?  I'll add some comments.  BTW I didn't modify this code at all, just refactored so I could use the same bidirectional LSTM code to preprocess control sequences.

adarob: yeah h is output -- but putting "(h)" will just make it more explicit  -- i think just adding that is enough

iansimon: Added that and a tiny bit more.

PR #449, opened 2020-05-08
add MusicVAE models for generating melodic rhythms and shapes directly

iansimon:


music/src/core/data.ts
```diff
@@ -370,8 +384,7 @@ export class MelodyConverter extends DataConverter {
     super(args);
     this.minPitch = args.minPitch;
     this.maxPitch = args.maxPitch;
-    this.ignorePolyphony =
-        (args.ignorePolyphony === undefined) ? true : args.ignorePolyphony;
+    this.ignorePolyphony = args.ignorePolyphony;
```
adarob, line 387: does this change the behavior?

music/src/core/data.ts
```diff
@@ -396,6 +409,143 @@ export class MelodyConverter extends DataConverter {
...
+    noteSequence.totalQuantizedSteps = rhythm.length;
+    return noteSequence;
+  }
+}
+
+/**
+ * Converts between a monophonic, quantized `NoteSequence` containing a melody
+ * and a `Tensor` representing only the *shape* of the melody.
+ *
+ * The shape is represented as a `numSteps`-by-3 `Tensor` containing a one-hot
```
adarob, line 494: I think "`[num_steps, 3]`-shaped `Tensor`" is more readable.

music/src/core/data.ts
```diff
@@ -396,6 +409,143 @@ export class MelodyConverter extends DataConverter {
...
+        this.numSteps);
+    return this.melodyControl.extract(melody);
+  }
+}
+
+/**
+ * Converts between a monophonic, quantized `NoteSequence` containing a melody
+ * and a `Tensor` representing only the *rhythm* of the melody.
+ *
+ * The rhythm is represented as a `numSteps`-by-1 `Tensor` with 1 in the
```
adarob, line 455: I think "`[num_steps, 1]`-shaped `Tensor`" is more readable.

music/src/music_vae/model.ts
```diff
@@ -1034,6 +1107,26 @@ class MusicVAE {
...
   }
 
+  /**
+   * Encodes the input `Tensor`s into latent vectors.
+   *
+   * @param inputTensors A 3D `Tensor` of sequences to encode.
+   * @returns A `Tensor` containing the batch of latent vectors, sized
+   * `[inputTensors.shape[0], zSize]`.
+   */
+  async encodeTensors(inputTensors: tf.Tensor3D) {
```
adarob, line 1117: can we call these new methods from the old ones to avoid duplication and make it more readable?

music/src/core/data.ts
iansimon: Done.

music/src/core/data.ts
iansimon: Done.

music/src/core/data.ts
iansimon: No, because the Melody code now uses true as the default value for ignorePolyphony.

music/src/music_vae/model.ts
iansimon: Talked offline, I don't want to support conditioning yet in the *Tensors methods.

PR #450, opened 2020-05-12
add melody register control to MusicVAE

iansimon:


iansimon:
@cannoneyed @cjcai and @notwaldorf may be interested...

music/src/core/melodies.ts
```diff
@@ -254,3 +254,65 @@ export class MelodyShape implements MelodyControl {
...
+  readonly boundaryPitches: number[];
+  readonly depth: number;
+
+  constructor(boundaryPitches: number[]) {
+    this.boundaryPitches = boundaryPitches;
+    this.depth = boundaryPitches.length + 1;
+  }
+
+  private meanMelodyPitch(melody: Melody): number {
+    // Compute mean melody pitch, weighted by time steps.
```
adarob, line 272: weighted by non-silent time steps

iansimon: Edited to clarify; I think "non-silent" is implied by "pitch" but I made it clearer that it's explicitly weighting each pitch by how long it is sustained.

PR #455, opened 2020-05-29
generate samples similar to a source sample in MusicVAE

iansimon:


iansimon:
@cjcai in case you're interested...

music/src/music_vae/model.ts
```diff
@@ -1355,7 +1355,7 @@ class MusicVAE {
     }
     const randZs: tf.Tensor2D =
         tf.tidy(() => tf.randomNormal([numSamples, this.decoder.zDims]));
-    const outputTensors = this.decodeTensors(randZs, temperature);
+    const outputTensors = await this.decodeTensors(randZs, temperature);
```
adarob, line 1358: did we not need this before or was that a bug?

music/src/music_vae/model.ts
```diff
@@ -1403,6 +1403,106 @@ class MusicVAE {
             'Sampling completed', startTime, 'MusicVAE', logging.Level.DEBUG));
     return outputSequences;
   }
+
+  /**
+   * Generates similar tensors to an input tensor.
```
adarob, line 1408: I wonder if this should be implemented in a downstream app rather than in the core API. @notwaldorf what do you think?  It seems a bit experimental/specialized.

adarob: Happy to be convinced!

music/src/music_vae/model.ts
iansimon: I think it could result in the below dispose of randZs happening before they get used, though I never observed that behavior.

music/src/music_vae/model.ts
iansimon: I agree with you, just not sure where to put it.  This functionality was requested by cjcai@

iansimon: @cjcai I mean

notwaldorf: It's true that it's not super great here, but copy pasting this code in any app that needs it also isn't great. Also, I feel the API ship has sailed a little bit because now MusicVAE has 3 functions that require tensors (the new controls stuff), so we can't even pretend like we haven't exposed tensor implementation anymore

iansimon: It's not so much about the tensors but the fact that I added this composite function to the MusicVAE API, when maybe it should be in some downstream library.

Though I suppose interpolate is also a composite function...

adarob: Yes, interpolate is composite but it's also a fairly standard use-case. This seems a bit more experimental. Also, couldn't you achieve the same result by combining a sample and interpolate?

iansimon: Not easily, because interpolate generates a fixed number of samples equally spaced between the endpoints.  You'd end up throwing most of them away.

cjcai: I don't have full context here, but "get similar" seems like a very standard use case at least in non-generative models (e.g. with just normal embedding spaces). I suspect this feature could get a lot of useage if integrated into musicVAE...

PR #458, opened 2020-06-02
add key conditioning to MusicVAE

iansimon:
Another one for @cjcai

### DIFF
--- a/music/demos/melody_controls.ts
+++ b/music/demos/melody_controls.ts
@@ -31,6 +31,8 @@ const MEL_SHAPE_CKPT = `${CHECKPOINTS_DIR}/music_vae/mel_shape`;
 
 const MEL_MULTI_DIR = `${CHECKPOINTS_DIR}/music_vae/mel_multicontrol`;
 const MEL_MULTI_CONTROLS = `${MEL_MULTI_DIR}/mel_2bar_multicontrol_tiny_fb16`;
+const MEL_MULTI_CONTROLS_KEY =
+    `${MEL_MULTI_DIR}/mel_2bar_multicontrol_key_tiny_fb16`;
 const MEL_MULTI_RHYTHM = `${MEL_MULTI_DIR}/mel_rhythm_2bar_tiny_fb16`;
 const MEL_MULTI_SHAPE = `${MEL_MULTI_DIR}/mel_shape_2bar_tiny_fb16`;
 
@@ -141,11 +143,14 @@ async function runMelShape() {
 
 async function runMelMulti() {
   const mvaeMel = new mm.MusicVAE(MEL_MULTI_CONTROLS);
+  const mvaeMelKey = new mm.MusicVAE(MEL_MULTI_CONTROLS_KEY);
   const mvaeRhythm = new mm.MusicVAE(MEL_MULTI_RHYTHM);
   const mvaeShape = new mm.MusicVAE(MEL_MULTI_SHAPE);
 
-  await Promise.all(
-      [mvaeMel.initialize(), mvaeRhythm.initialize(), mvaeShape.initialize()]);
+  await Promise.all([
+    mvaeMel.initialize(), mvaeMelKey.initialize(), mvaeRhythm.initialize(),
+    mvaeShape.initialize()
+  ]);
 
   let start = performance.now();
 
@@ -165,6 +170,19 @@ async function runMelMulti() {
 
   writeTimer('mel-multi-sample-time', start);
   writeNoteSeqs('mel-multi-seqs', [rhythmSeq, shapeSeq, melodySeqs[0]]);
+
+  const melodySeqsIonian = await mvaeMelKey.sample(
+      1, null, ['C'], undefined, undefined, controlTensor, 0);
+  const zKey =
+      await mvaeMelKey.encode(melodySeqsIonian, ['C'], controlTensor, 0);
+  const melodySeqsLydian = await mvaeMelKey.decode(
+      zKey, null, ['C'], undefined, undefined, controlTensor, 7);
+  const melodySeqsMixolydian = await mvaeMelKey.decode(
+      zKey, null, ['C'], undefined, undefined, controlTensor, 5);
+
+  writeNoteSeqs(
+      'mel-multi-key-seqs',
+      [melodySeqsIonian[0], melodySeqsLydian[0], melodySeqsMixolydian[0]]);
 
   start = performance.now();
 
--- a/music/demos/melody_controls.ts
+++ b/music/demos/melody_controls.ts
@@ -73,14 +73,16 @@ async function runMelControls() {
   const teapotControls = tf.concat([teapotRhythm, teapotShape], 1);
 
   let start = performance.now();
-  const interp = await mvae.interpolate(inputs, 5, null, ['A'], teapotControls);
+  const interp = await mvae.interpolate(
+      inputs, 5, null,
+      {chordProgression: ['A'], extraControls: teapotControls});
   writeTimer('mel-controls-interp-time', start);
   writeNoteSeqs('mel-controls-interp', interp);
 
   const sampleControls = tf.concat([DOTTED_RHYTHM, UDUDF_SHAPE], 1);
   start = performance.now();
-  const sample =
-      await mvae.sample(4, null, ['C'], undefined, undefined, sampleControls);
+  const sample = await mvae.sample(
+      4, null, {chordProgression: ['C'], extraControls: sampleControls});
   writeTimer('mel-controls-sample-time', start);
   writeNoteSeqs('mel-controls-samples', sample);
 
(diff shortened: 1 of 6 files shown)


iansimon:
Also may be of interest to @cannoneyed

music/src/music_vae/model.ts
```diff
@@ -999,14 +999,16 @@ class MusicVAE {
...
+   * (C major / A minor) and 11 (B major / G# minor).
    *
    * @returns An array of interpolation `NoteSequence` objects, as described
    * above.
    */
   async interpolate(
       inputSequences: INoteSequence[], numInterps: number|number[],
       temperature?: number, chordProgression?: string[],
-      extraControls?: tf.Tensor2D) {
+      extraControls?: tf.Tensor2D, key?: number) {
```
cannoneyed, line 1011: I'm a little confused about the use of `extraControls`, when combined with key. What are the conditioning signals that are typically passed into `extraControls`?

music/demos/melody_controls.ts
```diff
@@ -166,6 +171,19 @@ async function runMelMulti() {
   writeTimer('mel-multi-sample-time', start);
   writeNoteSeqs('mel-multi-seqs', [rhythmSeq, shapeSeq, melodySeqs[0]]);
 
+  const melodySeqsIonian = await mvaeMelKey.sample(
+      1, null, ['C'], undefined, undefined, controlTensor, 0);
+  const zKey =
+      await mvaeMelKey.encode(melodySeqsIonian, ['C'], controlTensor, 0);
+  const melodySeqsLydian = await mvaeMelKey.decode(
+      zKey, null, ['C'], undefined, undefined, controlTensor, 7);
+  const melodySeqsMixolydian = await mvaeMelKey.decode(
```
cannoneyed, line 180: Just to straighten out the conflation of chord / key... Is the thought process that the combination of 7 (which I thought I read to mean GMaj) _plus_ the CMaj expected to push the model to generate notes in MixoLydian? There's not an explicit mode conditioning signal, correct?

I'm thinking of the case where you'd want to retain the key but change the mode... would you do that by holding the `key` number the same while adjusting the chord progression conditioning signal?

music/src/music_vae/model.ts
iansimon: Here it's general, but I've been using it to pass melody rhythm, shape, and register.

music/demos/melody_controls.ts
iansimon: By retain the key but change the mode, do you mean e.g. playing in C major but switch the underlying chord from C (Ionian) to G (Mixolydian), or do you mean playing in a mode rooted at C but change from C Ionian to C Dorian etc.?  If the first, then yes you would retain key but change the chord.

music/src/music_vae/model.ts
iansimon: I feel a little bit bad about making the API less clean @adarob; if this is really bothersome I could clean it up a bit here and have some higher level wrapper class for doing a bunch of the melody experiments we want to try.  But either way I'd have to expose extraControls (or something a lot like it) here...

iansimon:
> I'd love to know a bit more how this conditioning signal is generated. I'm expecting a fair bit of ambiguity around the interplay between the chord and key, but I'm interested in playing around with it soon!

What do you mean by "how this conditioning signal is generated"?  Do you mean what do the tensors passed to the model look like?  Currently it's a one-hot for chord with 49 values (4 triad types * 12 pitch classes + no-chord), and another one-hot for key with 12 values.

music/demos/melody_controls.ts
cannoneyed: I'm more thinking about the latter... From a UX perspective, I'm imagining that it might be fun for the user to switch modes "on-the-fly", and in my mind that operation keeps the root note the same (switch from C Ionian to C Aeolian). It's pretty fun to swap relative-encoded melodies between modes.

In this case do you just do something like this? 

```typescript
decode(zKey, null, ['C'], undefined, undefined, controlTensor, 0) // C Ionian
decode(zKey, null, ['Cmin'], undefined, undefined, controlTensor, 0) // C Aeolian

or is it

decode(zKey, null, ['C'], undefined, undefined, controlTensor, 0) // C Ionian
decode(zKey, null, ['Cmin'], undefined, undefined, controlTensor, 9) // C Aeolian
```

cannoneyed:
> What do you mean by "how this conditioning signal is generated"? Do you mean what do the tensors passed to the model look like? Currently it's a one-hot for chord with 49 values (4 triad types * 12 pitch classes + no-chord), and another one-hot for key with 12 values.

Yeah, this is mostly what I'm looking for... I'm a bit curious how far those chord one-hots extend, and whether or not there's some interesting things you could do with other encoding schemes (eg one-hot vectors for relative notes of the chord (something like `onehot(0, 4, 7)` for `I`)... just a thought.

music/demos/melody_controls.ts
iansimon: It would be neither of those :)  You'd pass 0 as the key the first time and 3 (key of Eb) the second time.

music/src/music_vae/model.ts
adarob: Ideally we would make all of these optional arguments part of an interface that can be more easily extended. For example, we would have something like:

interface ConditioningArgs {
  chordProgression?: string[];
  extraControls?: tf.Tensor2D;
  key?: number;
}

This would be a breaking change for apps using chordProgression or extraControls, but we can do a full version number bump. Thoughts @notwaldorf ?

notwaldorf: I'd be happy with a full version bump, I'd like to remove some old code from the visualizer while we're at it too. We've been afraid of major bumping for too long

notwaldorf: like i feel 1.x -> 2.x is really scary, but like 4.x -> 5.x is peanuts

iansimon: Okay, I'll try to clean up the conditioning code.

iansimon: What do you think of the changes I just made?

music/demos/melody_controls.ts
cannoneyed: Sorry, maybe we're not on the same page here but I'm confused. How would you express switching the conditioning of a melody from C Major to C minor?

cannoneyed: Nevermind, I see it now with the comment for `MusicVAEControlArgs`... there's no other modal conditioning, though, correct?

music/src/music_vae/model.ts
```diff
@@ -669,6 +669,21 @@ export interface MusicVAESpec {
...
+ * @property chordProgression (Optional) Chord progression, an array of chord
+ * symbol strings like "Cmaj7", "Abdim", "C#m7", etc.
+ * @property key (Optional) Key, an integer between 0 (C major / A minor) and 11
+ * (B major / G# minor).
+ * @property extraControls (Optional) Additional control tensors.
+ */
+export interface MusicVAEControlArgs {
+  chordProgression?: string[];
+  key?: number;
+  extraControls?: tf.Tensor2D;
```
cannoneyed, line 684: Just curious why `extraControls` is simply a concatenated `Tensor2D` of rhythm/contour and not specific a key/value for each of those?

iansimon: Do you mean why is `extraControls` a `Tensor2D` instead of a dict of `name` -> `Tensor2D`?  And then the model config can specify which controls are needed?  That's not a bad idea...

music/demos/melody_controls.ts
iansimon: Correct, there's no mode, just chord and key.

music/src/music_vae/model.ts
adarob: This looks much better to me!!

music/src/music_vae/model.ts
iansimon: @adarob What do you think about this?

adarob: I like @cannoneyed's suggestion if it's easy to do.

adarob: But also, why are these tensors not just separate fields in the interface?

iansimon: I think I don't want to define controls like rhythm or shape contour here as they are (as we've been using them) specific to melody.  But I can make `extraControls` a dict.

iansimon: Done, and updated checkpoints to work with the new (and old) code.

music/src/music_vae/model.ts
cannoneyed: Agreed. This is looking good

notwaldorf:
0h don't wait on me friends, lgtm

PR #469, opened 2020-06-30
More CSS-friendly SVG visualizers

cifkao:
Adding `class="note"` and a `--midi-velocity` CSS variable.

This is now possible:
```css
svg rect.note {
  fill: #28f;
  opacity: calc(var(--midi-velocity) / 127);
}
```
image

Or even:
```css
svg rect.note {
  fill: rgb(calc((var(--midi-velocity) - 64) * 4), 30, calc(255 - var(--midi-velocity)));
}
```
image

### DIFF
--- a/music/src/core/visualizer.ts
+++ b/music/src/core/visualizer.ts
@@ -339,9 +339,10 @@ export class Visualizer extends PianoRollCanvasVisualizer {
 }
 
 /**
- * HTML data attribute key-value pair.
+ * HTML/CSS key-value pairs.
  */
 type DataAttribute = [string, any];  // tslint:disable-line:no-any
+type CSSProperty = [string, string | null];
 
 /**
  * Abstract base class for a `NoteSequence` visualizer.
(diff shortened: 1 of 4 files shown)


music/src/core/visualizer.ts
```diff
@@ -438,8 +439,13 @@ export abstract class BaseSVGVisualizer extends BaseVisualizer {
         ['isDrum', note.isDrum === true],
         ['pitch', note.pitch],
       ];
+      const cssProperties: CSSProperty[] = [
+        ['--midi-velocity',
```
notwaldorf, line 443: Just to triple check I understand how this works: this property will be set on each `rect.note`, and will be different for each rect, right? the velocity is definitely different for each note, so I want to make sure that is reflected

notwaldorf:
(omg YES rainbow gradients, i love this)

music/src/core/visualizer.ts
cifkao: Yes, exactly. Every `rect` will basically get a `style="--midi-velocity: ..."`.

notwaldorf:
Now that I think of it, I think it would be super neat if you added a section to the visualizer demo with your rainbow gradient and a little blurb about how some CSS classes are available for styling. What do you think? @cifkao

notwaldorf:
(sorry to make you do extra work!)

cifkao:
Oookay, here you go!

BTW, in the future, it should be possible to access any HTML properties as in `attr(data-velocity)` so we wouldn't need the custom CSS properties, but currently it's not supported anywhere.

PR #484, opened 2020-07-24
Bind fetch to window to eliminate weird fetch issue

cannoneyed:
The ES5 build was throwing the following error when trying to use the `window.fetch` method from the new compat code: 

```
Failed to execute 'fetch' on 'Window': Illegal invocation
```

This is because webpack borks the context of the `fetch` method that's exported from the global compat file, so we need to bind it to `window` before exporting it to have it work properly :/

### DIFF
--- a/music/src/core/compat/global_browser.ts
+++ b/music/src/core/compat/global_browser.ts
@@ -18,6 +18,6 @@
  * limitations under the License.
  */
 
-export const fetch = window.fetch;
+export const fetch = window.fetch.bind(window);
 export const performance = window.performance;
 export const navigator = window.navigator;
--- a/music/src/core/compat/global.ts
+++ b/music/src/core/compat/global.ts
@@ -28,6 +28,6 @@ export interface Performance {
 }
 
 // tslint:disable:no-require-imports
-export const fetch = isNode ? require('node-fetch') as (typeof window.fetch) :  window.fetch;
+export const fetch: typeof window.fetch = isNode ? require('node-fetch') : window.fetch.bind(window);
 export const performance: Performance = isNode ? require('./performance_node') : window.performance;
 export const navigator = isNode ? require('./navigator_node') : window.navigator;


notwaldorf:
Do you have any advice on how we can write tests for this? We're not really testing that the node/es5 bundles really work> You don't have to do it, it's just something to think about

music/src/core/compat/global_browser.ts
```diff
@@ -18,6 +18,6 @@
  * limitations under the License.
  */
 
-export const fetch = window.fetch;
+export const fetch = window.fetch.bind(window);
 export const performance = window.performance;
```
notwaldorf, line 22: we don't have to do this for performance? just double checking

cannoneyed: no, it's a (dumb) JS context thing - since fetch is a method and we export it without binding it, it's messing with the internals of the method (as it's somehow being invoked with a context that's *not* the `window` object, probably because of the way webpack is making a module object)

`performance` and `navigator` are each objects, which have the rest of their methods and properties attached to them, so calling a method on them will have the correct context.

cannoneyed: * waves hands *

cannoneyed:
Yeah, I'm not sure what the best way to test all of this is, aside from a headless browser test runner... At Patreon we used [nightmare.js](https://github.com/segmentio/nightmare) for our browser "integration tests", and it was actually pretty good. Way easier to set up than protractor or something crazy like browserstack

notwaldorf:
Actually, I think I'm being dumb. We can write a simple puppeteer test that tries to open some html files that basically put some text out if everything is loaded correctly, and since those are headless we don't have to interact with any VMs on Travis or anything

music/src/core/compat/global_browser.ts
notwaldorf: ah yeah, duh you're right

notwaldorf:
published as 1.18.1

notwaldorf:
Annnnd confirming that if I pin https://hello-magenta.glitch.me/ to 1.18.1 it works again! Thanks so much Andy!

PR #486, opened 2020-07-28
Add global object helpers

cannoneyed:
In order to keep fighting the fight of browser/worker/node compatibility of Magenta.js, we need to properly ensure that global objects are provided correctly across the different build environments. Earlier changes had neglected to deal with web workers correctly, so this PR adds a few helper methods to ensure that the global/window/self global object is being used properly in the given setting.

It also adds a few shim methods to help with the OfflineAudioContext global, which will throw more helpful errors when attempting to be used in settings without the context (worker and node).

In the near future, we'll need to set up puppeteer headless browser tests to ensure that all of this is working correctly in all of the different environments that magenta is being used in.

### DIFF
--- a/music/src/core/audio_utils.ts
+++ b/music/src/core/audio_utils.ts
@@ -23,20 +23,11 @@ import * as ndarray from 'ndarray';
 //@ts-ignore
 import * as resample from 'ndarray-resample';
 
-import {fetch} from '../core/compat/global';
+import { fetch, getOfflineAudioContext, isSafari } from '../core/compat/global';
 import * as logging from './logging';
 
-// Safari Webkit only supports 44.1kHz audio.
-const WEBKIT_SAMPLE_RATE = 44100;
 const SAMPLE_RATE = 16000;
-// tslint:disable-next-line:no-any
-const appeaseTsLintWindow = (window as any);
-const isSafari = appeaseTsLintWindow.webkitOfflineAudioContext as boolean;
-// tslint:disable-next-line:variable-name
-const offlineCtx = isSafari ?
-    new appeaseTsLintWindow.webkitOfflineAudioContext(
-        1, WEBKIT_SAMPLE_RATE, WEBKIT_SAMPLE_RATE) :
-    new appeaseTsLintWindow.OfflineAudioContext(1, SAMPLE_RATE, SAMPLE_RATE);
+const offlineCtx = getOfflineAudioContext(SAMPLE_RATE);
 
 /**
  * Parameters for computing a spectrogram from audio.
@@ -63,8 +54,8 @@ export interface SpecParams {
  */
 export async function loadAudioFromUrl(url: string): Promise<AudioBuffer> {
   return fetch(url)
-      .then(body => body.arrayBuffer())
-      .then(buffer => offlineCtx.decodeAudioData(buffer));
+    .then(body => body.arrayBuffer())
+    .then(buffer => offlineCtx.decodeAudioData(buffer));
 }
 
 /**
@@ -92,7 +83,7 @@ export async function loadAudioFromFile(blob: Blob): Promise<AudioBuffer> {
 }
 
 export function melSpectrogram(
-    y: Float32Array, params: SpecParams): Float32Array[] {
+  y: Float32Array, params: SpecParams): Float32Array[] {
   if (!params.power) {
     params.power = 2.0;
   }
@@ -147,7 +138,7 @@ function getMonoAudio(audioBuffer: AudioBuffer) {
   }
   if (audioBuffer.numberOfChannels !== 2) {
     throw Error(
-        `${audioBuffer.numberOfChannels} channel audio is not supported.`);
+      `${audioBuffer.numberOfChannels} channel audio is not supported.`);
   }
   const ch0 = audioBuffer.getChannelData(0);
   const ch1 = audioBuffer.getChannelData(1);
@@ -160,7 +151,7 @@ function getMonoAudio(audioBuffer: AudioBuffer) {
 }
 
 export async function resampleAndMakeMono(
-    audioBuffer: AudioBuffer, targetSr = SAMPLE_RATE) {
+  audioBuffer: AudioBuffer, targetSr = SAMPLE_RATE) {
   if (audioBuffer.sampleRate === targetSr) {
     return getMonoAudio(audioBuffer);
   }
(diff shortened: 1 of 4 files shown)


music/src/core/compat/global_browser.ts
```diff
@@ -18,6 +18,37 @@
  * limitations under the License.
  */
 
-export const fetch = window.fetch.bind(window);
-export const performance = window.performance;
-export const navigator = window.navigator;
+function getGlobalObject() {
+  if (typeof globalThis !== 'undefined') { return globalThis; }
```
notwaldorf, line 22: Would love some comments about which exists where (self in workers, window in browsers, etc). I also don't know where `globalThis` comes from...whose sorcery is it? :D

music/src/core/compat/global.ts
```diff
@@ -31,3 +37,18 @@ export interface Performance {
 export const fetch: typeof window.fetch = isNode ? require('node-fetch') : window.fetch.bind(window);
 export const performance: Performance = isNode ? require('./performance_node') : window.performance;
 export const navigator = isNode ? require('./navigator_node') : window.navigator;
+
+export function isSafari() {
+  if (isNode) {
+    throw new Error('Cannot check if Safari in Node.js');
+  }
+  // tslint:disable-next-line:no-any
+  return (window as any).safari !== undefined;
```
notwaldorf, line 46: whoaaaa where is this getting set? i think if we're setting this it's not super great, because it's a very vague noun on the window object so there is a chance in the future someone can accidentally break this

notwaldorf:
Thanks for fighting the good fight Andy!!!

music/src/core/compat/global_browser.ts
cannoneyed: Yeah, I'll add that in! `globalThis` is neat - I just learned about it while researching the problem. It's the new cross-browser way to consistently access the global object from both browser/worker contexts (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis)

music/src/core/compat/global.ts
cannoneyed: No, this is another sweet thing that I didn't know existed until I did some research. Seems like a simpler way to check than parsing `navigator.userAgent`

notwaldorf: Hmmmmm this isn’t in the spec (https://developer.mozilla.org/en-US/docs/Web/API/Window). Will it be true on chrome on iOS, or other webview on iOS? I really don’t know who is setting this and when, and I don’t trust it, I’m sorry. Can we revert to the tried and true navigator.userAgent please?

music/src/core/compat/global.ts
```diff
@@ -31,3 +37,17 @@ export interface Performance {
...
+    throw new Error('Cannot check if Safari in Node.js');
+  }
+  return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+}
+
+export function getOfflineAudioContext(sampleRate: number): OfflineAudioContext {
+  if (isNode) {
+    throw new Error('Cannot use offline audio context in Node.js');
+  }
+  return new window.OfflineAudioContext(1, sampleRate, sampleRate);
```
notwaldorf, line 52: I think this needs to be the if Safari? webkitOfflineAudioContext : offlineAudioContext, since you could be realistically developing and testing on Safari

music/src/core/compat/global_browser.ts
```diff
@@ -18,6 +18,42 @@
...
+}
+
+export function getOfflineAudioContext(sampleRate: number): OfflineAudioContext {
+  if (isWorker()) {
+    throw new Error('Cannot use offline audio context in a web worker.');
+  }
+
+  // tslint:disable-next-line:no-any variable-name
+  const SafariOfflineCtx = (globalObject as any).webkitOfflineAudioContext;
+  return isSafari() ? new SafariOfflineCtx(1, sampleRate, sampleRate) :
```
notwaldorf, line 57: This check is technically incorrect, but if you've tested it in Safari/Chrome on iOS etc, firefox on iOS etc it should be fine. IMO you should be feature checking whether or not `webkitOfflineAudioContext` exists, since that's what you're trying to use, rather than using the user agent string at all. I think the best practices around user agent strings is "use them if you can't find a feature to check", which in this case we have

music/src/core/compat/global_node.ts
```diff
@@ -22,3 +22,11 @@
 export const fetch = require('node-fetch');
 export const performance = require('./performance_node');
 export const navigator = require('./navigator_node');
+
+export function isSafari(): boolean {
```
notwaldorf, line 26: Should it throw or should it return false, since Node isn't Safari?

music/src/core/audio_utils.ts
```diff
@@ -23,20 +23,14 @@ import * as ndarray from 'ndarray';
...
 const SAMPLE_RATE = 16000;
-// tslint:disable-next-line:no-any
-const appeaseTsLintWindow = (window as any);
-const isSafari = appeaseTsLintWindow.webkitOfflineAudioContext as boolean;
-// tslint:disable-next-line:variable-name
-const offlineCtx = isSafari ?
-    new appeaseTsLintWindow.webkitOfflineAudioContext(
-        1, WEBKIT_SAMPLE_RATE, WEBKIT_SAMPLE_RATE) :
-    new appeaseTsLintWindow.OfflineAudioContext(1, SAMPLE_RATE, SAMPLE_RATE);
+const sampleRate = isSafari() ? WEBKIT_SAMPLE_RATE : SAMPLE_RATE;
```
notwaldorf, line 32: i feel we don't use this `sampleRate` anywhere else other than this constructor, and we don't use `isSafari` outside of the offlineCtx constructor, so i would maybe not export isSafari and just pass the two sample rates to getOfflineAudioContext instead

music/src/core/compat/global.ts
cannoneyed: Yeah, I thin that's right

music/src/core/compat/global_browser.ts
cannoneyed: Yup

music/src/core/audio_utils.ts
cannoneyed: Yeah, that makes sense

music/src/core/compat/global_node.ts
cannoneyed: This is a tricky question - I think it's better to throw early, because checking if it's Safari is kind of meaningless in Node. But I'll make it so it just returns false... There'll be an error if people try to use audio_utils related stuff in Node, which I think is fine (eg we don't have to go far out of our way to prevent people from trying that)

music/src/core/audio_utils.ts
cannoneyed: I also updated `isSafari` and `isWorker` to just export boolean values, since exporting a truthy function is a bit dangerous.

music/src/core/compat/global.ts
notwaldorf: i dont think this change went through?

cannoneyed: Yup, you're right

music/src/core/compat/global.ts
```diff
@@ -31,3 +37,5 @@ export interface Performance {
 export const fetch: typeof window.fetch = isNode ? require('node-fetch') : window.fetch.bind(window);
 export const performance: Performance = isNode ? require('./performance_node') : window.performance;
 export const navigator = isNode ? require('./navigator_node') : window.navigator;
+
+export { isSafari, getOfflineAudioContext } from './global_browser';
```
notwaldorf, line 41: yesss love this

PR #487, opened 2020-07-28
support non-hierarchical split decoders in MusicVAE

iansimon:


iansimon:
The model I've been using for testing isn't one we should release.  I'm training a new one now but can I merge this first so @cjcai and @cannoneyed can use it?

notwaldorf:
Doing a release!

notwaldorf:
released as 1.19.0

PR #495, opened 2020-08-12
Fixed bug with Coconet when running with tfjs-node backend

divisiondeariza:
There was a weird bug when you use coconet with tfjs-node backend, it required [here](https://github.com/magenta/magenta-js/blob/b149ce62e7b58075e7177d3b33be123fb6706144/music/src/coconet/model.ts#L518) the samples to be casted to float32.

Now it works well with default and node backends.

### DIFF
--- a/music/src/coconet/model.ts
+++ b/music/src/coconet/model.ts
@@ -565,7 +565,8 @@ class Coconet {
             predictions.shape[0], predictions.shape[1], predictions.shape[3],
             predictions.shape[2]
           ])
-          .transpose([0, 1, 3, 2]);
+          .transpose([0, 1, 3, 2])
+          .cast("float32");
     });
   }
 }


divisiondeariza:
Warning: This fix make the model "run" with tfjs-node backend but it produces a very weird output, doesn't sound as with webgl backend at all!

notwaldorf:
If it produces very weird output that probably means all the values are incorrect, i'm afraid. can you be more clear about what doesn't work on node with the current version of Coconet?

divisiondeariza:
Hi

Well, I tried to generate a sequence with coconet model using tfjs-node. [there is a snipped of what i tried to do in a backend in express](https://github.com/divisiondeariza/magenta-rest/blob/master/routes/coconet.js)

When I tried with purejs backend (say, removing [this line](https://github.com/divisiondeariza/magenta-rest/blob/master/routes/coconet.js#L12)) It works (but it's very slow).

I had to do the change in this pull request in order to get the model work with tfjs-node backend (otherwise it throws an error because it requires the samples to be casted to float32).

With the change i made it works well with purejs backend and webgl backend, with tfjs-node it runs but the sequences that it generates sound bad, you can notice the difference, it's just not harmonic as with other backends.

[I read here ](https://github.com/tensorflow/tfjs/issues/1112#issuecomment-459990325) that webgl and tfjs-node backends may produce "slightly different probability outputs". I tried to fix that but I didn't got any luck.

notwaldorf:
I am very worried about merging this PR. It looks like while this _may_ fix some broken behaviour on tfjs-node, it definitely introduces more broken behaviour -- the results are definitely bad, and in my opinion broken weights is just as bad as the model not working at all. I think the proper solution (like we've done for other models) is to scale the actual weights during training, but unfortunately I don't think anyone has the bandwidth for that right now. 

If you want to use this fix locally (though it sounds like you'd get very bad samples), I would recommend taking the weights in the checkpoint and scaling _those_ locally, and then using the model code as is.

notwaldorf:
I'm tempted to close this PR. WDYT @adarob?

PR #497, opened 2020-08-20
Fix typo for "In the browser" example and replace Codepen link with a Glitch project.

vibertthio:
The original demo on CodePen is much more complicated than code snippet in "In the browser" example. I make a new one on Glitch similar to the usage with the example.


notwaldorf:
I think that we do want to keep the link to endless trios, since it's really cool. If there's a problem with the code, @adarob owns the pen and can fix it.

vibertthio:
Got it. I restore the CodePen link and list both links under as bullet points under the code snippet. Does that make sense?

PR #498, opened 2020-08-22
add TrioRhythmConverter and support controlArgs in MusicVAE tensor methods

iansimon:


music/src/core/data.ts
```diff
@@ -460,7 +466,6 @@ abstract class MelodyControlConverter extends DataConverter {
  * `toNoteSequence` returns a `NoteSequence` with drum hits at the note-on
  * steps.
  */
-export interface MelodyRhythmConverterArgs extends BaseConverterArgs {}
```
adarob, line 463: why remove these?

iansimon: It was actually never being used, the MelodyRhythmConverter just uses the MelodyConverterArgs.

PR #499, opened 2020-08-23
Add core/tone_io.ts to make customizing Tone.js sounds easier

vibertthio:
1. `noteSequenceToToneEvents` and `toneEventsToNoteSequence` in `tone_io.ts`
2. A basic round trip test for the conversion
3. A demo page for basic usage. Check out this [glitch](https://hello-magenta-3.glitch.me/) to see the usage for real

This PR is my first attempt to "making magenta-js more interactive" 😋
Any suggestions/comments/feedbacks are super welcomed!


notwaldorf:
@cannoneyed Andy, do you want to take a look at this? I think you have the most Tone opinions around :)

vibertthio:
@cannoneyed  Thanks for the suggestions! After sitting on this for a few months, I think this solution is not ideal as a stable API. I think building a separate helper library for this use case is a better solution because both ends are still volatile (magenta.js and tone.js). I will close this until I come up with a better solution!

PR #513, opened 2020-11-06
feature(ddsp):

EdwinToh:
- initial commit of DDSP integration
- added methods for getAudioFeatures and toneTransfer
- added demo
- added models to repo for now, will move to checkpoints


notwaldorf:
FYI we should upload the models to a cloud storage bucket rather than checking them in the repo itself. I can help with that if needed

jesseengel:
The checkpoints are up at https://storage.googleapis.com/magentadata/js/checkpoints/ddsp
for example: https://storage.googleapis.com/magentadata/js/checkpoints/ddsp/flute/model.json

music/demos/ddsp.ts
```diff
@@ -0,0 +1,176 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MODEL } from '../src/ddsp/constants';
+import * as mm from '../src/index';
+
+function floatTo16BitPCM(output, offset, input) {
```
jesseengel, line 21: It seems like some of these functions would go good in a general "audio utils" library. What do you think @notwaldorf @adarob ?

music/src/ddsp/audio_utils.ts
```diff
@@ -0,0 +1,117 @@
+/**
```
jesseengel, line 1: Yah, there's nothing ddsp specific about these. It might be nice to centralize all the audio_utils in the repo if that makes sense.

music/src/ddsp/loudness_utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    .conv1d(
+      sq as tf.Tensor2D,
+      tf.ones([frameSize, 1, 1]).div(frameSize) as Tensor3D,
+      hopSize,
+      'same'
+    )
+    .sqrt()
+    .squeeze();
+
+  // Amplitude to Db
```
jesseengel, line 48: It might be nice to split this out as a separate amplitude_to_db() function in an audio util library.

music/src/ddsp/loudness_utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  const hopSize = Math.floor(MODEL_SAMPLE_RATE / frameRate);
+
+  const audioTensor = tf.tensor1d(audioChannelData, 'float32');
+  const newSamplesCount = audioChannelData.length;
+
+  if (audioTensor === null) {
+    return [];
+  }
+
+  // Compute RMS energy
```
jesseengel, line 36: Same with this compute_rms_power() function

That would then help isolate this wrapper that does shifting and clipping on top of these two more fundamental functions.

music/src/ddsp/model.ts
```diff
@@ -0,0 +1,110 @@
...
+   * Synthesizes audio features based on model.
+   *
+   * @param inputAudioBuffer An AudioBuffer
+   * @returns Audio Features of provided audio buffer
+   */
+  async synthesize(
+    model: MODEL,
+    audioFeatures: AudioFeatures
+  ): Promise<Float32Array> {
+    return await toneTransfer(model, audioFeatures);
```
jesseengel, line 106: not to be picky about this, but technically toneTransfer() is the process of both extracting the audio features and then synthesizing.

You could rename toneTransfer to synthesize, and then have another convenience function toneTransfer that just takes in audio and returns the tone transferred audio. 

Also, I'm not sure it makes sense to have the synthesize function (toneTransfer() at the moment) as a separate method, because it always needs to be paired with a model, so it might make sense to just have it defined in the class itself.

music/src/ddsp/pitch_utils.ts
```diff
@@ -0,0 +1,191 @@
...
+  CONF_THRESHOLD,
+  PT_SLOPE,
+  PT_OFFSET,
+  PITCH_CONF_JITTER,
+} from './constants';
+import { AudioData } from './interfaces';
+import { Tensor } from '@tensorflow/tfjs';
+
+function upsample_linear(
+  spicePitchesOutput: number[],
```
jesseengel, line 29: you could name these a bit more generally, because it's actually just a linear interpolation function right?

music/src/ddsp/pitch_utils.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function getPitchHz(modelPitch: number) {
+  const fmin = 10.0;
+  const binsPerOctave = 12.0;
+  const cqtBin = modelPitch * PT_SLOPE + PT_OFFSET;
+  return fmin * Math.pow(2.0, (1.0 * cqtBin) / binsPerOctave);
+}
+
+async function getPitches(spiceModel: tf.GraphModel, inputData: AudioData) {
```
jesseengel, line 89: since this is so bound to the specifics of the spiceModel, it might make sense for us to make our own SpiceModel class that has this function and all these SPICE constants as attributes.

music/src/ddsp/spice.ts
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as tf from '@tensorflow/tfjs';
+import { SPICE_MODEL_URL } from './constants';
+
+async function startSpice() {
```
jesseengel, line 21: Yah, if you follow the same general model class API that the rest of magenta.js uses for the spice model as well, I think it would be clearer how to use it and more modular.

music/src/ddsp/model.ts
```diff
@@ -0,0 +1,110 @@
...
+
+class DDSP {
+  private initialized: boolean;
+
+  /**
+   * `DDSP` constructor.
+   *
+   * @param model The model to use
+   */
+  constructor() {
```
jesseengel, line 38: If we're following the model API for the rest of magenta.js I think it makes sense to provide the constructor with the url and constants, for example: https://github.com/magenta/magenta-js/blob/master/music/src/gansynth/model.ts#L30

jesseengel (commented):
Looks pretty good Edwin, awesome job!

I just left a couple comments on how I think restructuring to follow the common api in magenta.js for models (for both DDSP and SPICE) could help make things clearer and more modular.

Also, (less pressing) as a broader comment, it would definitely be nice to incorporate all the non ddsp specific audio_util functions into https://github.com/magenta/magenta-js/blob/master/music/src/core/audio_utils.ts (but that goes for other models too, so that could be a refactor later).

music/package.json
```diff
@@ -8,14 +8,15 @@
   "jsdelivr": "dist/magentamusic.js",
   "unpkg": "dist/magentamusic.js",
   "dependencies": {
-    "@tensorflow/tfjs": "^1.1.2",
+    "@tensorflow/tfjs": "^2.7.0",
```
notwaldorf, line 11: have you double checked everything else works when we switch to 2.x? Like, are all the tests passing, etc?

music/report.20201105.152937.93397.0.001.json
```diff
@@ -0,0 +1,689 @@
+
```
notwaldorf, line 1: I don't know what this file is but it probably shouldn't be checked in

music/src/ddsp/add_reverb.ts
```diff
@@ -0,0 +1,59 @@
+import * as Tone from 'tone';
```
notwaldorf, line 1: needs a license

music/src/ddsp/buffer_utils.ts
```diff
@@ -0,0 +1,31 @@
+export const arrayBufferToAudioBuffer = (
```
notwaldorf, line 1: needs a license

music/src/ddsp/constants.ts
```diff
@@ -0,0 +1,68 @@
+export const MIN_VRAM = 50; // used for memory check
```
notwaldorf, line 1: needs a licence

music/src/ddsp/interfaces.d.ts
```diff
@@ -0,0 +1,18 @@
+export interface AudioFeatures {
```
notwaldorf, line 1: needs a licence

music/src/ddsp/model.ts
```diff
@@ -0,0 +1,110 @@
...
+    tf.env().set('WEBGL_PACK', false);
+    tf.env().set('WEBGL_CONV_IM2COL', false);
+    tf.env().set('WEBGL_DELETE_TEXTURE_THRESHOLD', 100 * 1024 * 1024);
+  }
+
+  /**
+   * Disposes of any untracked `Tensors` to avoid GPU memory leaks.
+   */
+  dispose() {
+    if (!this.initialized) {
```
notwaldorf, line 61: we really don't have to dispose of anything here?

music/src/ddsp/pitch_utils.ts
```diff
@@ -0,0 +1,191 @@
...
+        // @ts-ignore
+        const jitter = 1.0 - noise.dataSync();
+        spicePitchesOutput.push(lastPitch * jitter);
+        noise.dispose();
+      }
+    }
+  }
+
+  // @ts-ignore
+  output[0].dispose();
```
notwaldorf, line 183: Why are there so many ts-ignores? I don't think normally we need to ignore the tensor disposing

music/src/ddsp/add_reverb.ts
```diff
@@ -0,0 +1,59 @@
...
+    }).connect(filter);
+    player.start();
+    // @ts-ignore
+    return reverb.ready;
+  }, arrayBuffer.length / sampleRate);
+  bufferWithReverb = await renderingPromise;
+  const bufferWithReverbData = bufferWithReverb.getChannelData(0);
+  // manual cleanup for safari
+  resampledAudioBuffer = null;
+  // @ts-ignore
```
notwaldorf, line 47: Hmm, why are there so many ignores?

music/src/ddsp/audio_utils.ts
```diff
@@ -0,0 +1,117 @@
...
+    tf.div(tf.log(440.0), tf.log(2))
+  );
+  frequenciesTensor = tf.mul(12, frequenciesTensor);
+  frequenciesTensor = tf.add(frequenciesTensor, 69);
+  return frequenciesTensor;
+}
+
+function shiftF0(f0Hz: number[], f0OctaveShift = 0.0) {
+  return tf.tidy(
+    // @ts-ignore
```
notwaldorf, line 59: also here

notwaldorf (changes requested):
Some initial cleanup comments, I haven't actually looked at any of the code yet

music/src/ddsp/add_reverb.ts
EdwinToh: @notwaldorf I have a TODO to fix types (which I privately DM'ed Jesse) so apologies for not stating it in this draft PR!  (It's also why this is still a draft PR and not ready for a review)

In this case, I'm actually struggling a little with TFJS and TS where they expect a Tensor in the return of a prediction but our models actually return an array with 2 tensors.

music/src/ddsp/model.ts
EdwinToh: I thought a little bit about this and decided to do it this way because of "getAudioFeatures" is actually agnostic of the models, and if you see the demo, you can see that you just need to run getAudioFeatures once, and then run synthesize with all 4 different models. WDYT?

music/package.json
EdwinToh: Not yet, but definitely something I am planning to do.

music/src/ddsp/pitch_utils.ts
EdwinToh: @notwaldorf I have a TODO to fix types (which I privately DM'ed Jesse) so apologies for not stating it in this draft PR! (It's also why this is still a draft PR and not ready for a review)

In this case, I'm actually struggling a little with TFJS and TS where they expect a Tensor in the return of a prediction but our models actually return an array with 2 tensors.

music/src/ddsp/model.ts
jesseengel: That's fine, but then the method should be called `synthesize` because that's what it does :). I'm just saying if someone makes a demo with only one model and wants a convenience function that goes straight audio2audio then a convenience function named toneTransfer would make sense.

music/demos/ddsp.ts
EdwinToh: this is a function used in the demo actually, I don't think the library should try to handle this but I'll leave it up to you guys to decide.

EdwinToh:
@jesseengel @notwaldorf This PR is now ready for review. I've addressed all of the comments, except for moving some audio functions into core.

While I wait for your review, I will be testing (and fixing) the other demoes/libraries to make sure that the upgrade to TFJS and Tone.js will not break them.

EdwinToh:
i've ran tests and manually tested all the demoes, everything is working. Just FYI, there're a few pre-existing type errors on Music Vae, Core, and Piano Genie that I did not fix since I didn't want this PR to touch those files.

music/demos/ddsp_tone-transfer.html
```diff
@@ -0,0 +1,92 @@
+<!-- Copyright 2020 Google Inc. All Rights Reserved.
```
jesseengel, line 1: super minor nit, but can you keep all snake case for the filename?

music/demos/ddsp_tone-transfer.ts
```diff
@@ -0,0 +1,209 @@
+/**
```
jesseengel, line 1: same here

music/demos/ddsp_tone-transfer.ts
```diff
@@ -0,0 +1,209 @@
...
+
+import { MODEL } from '../src/ddsp/constants';
+import * as mm from '../src/index';
+import { SPICE } from '../src/index';
+import { AudioFeatures } from '../src/ddsp/interfaces';
+
+export const MODEL_URL =
+  'https://storage.googleapis.com/magentadata/js/checkpoints/ddsp';
+
+export const PRESET_MODELS = {
```
jesseengel, line 26: Another super minor nit, but these are PRESET_MODEL_URLS technically right?

music/demos/index.html
```diff
@@ -55,6 +55,7 @@ <h4>
         <li><a href="piano_genie_conditioned.html">Play Piano Genie with specified key signatures and/or chord progressions</a></li>
         <li><a href="gansynth.html">Neural audio synthesis with GANSynth</a></li>
         <li><a href="coconet.html">Infill Bach-style 4-part harmonies with Coconet</a></li>
+        <li><a href="ddsp_tone-transfer.html">Transform audio into other musical instruments using DDSP</a></li>
```
jesseengel, line 58: same here

music/src/ddsp/model.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Synthesizes audio features based on model.
+   *
+   * @param inputAudioBuffer An AudioBuffer
+   * @returns Audio Features of provided audio buffer
+   */
+  async synthesize(
+    audioFeatures: AudioFeatures,
+    options?: ModelValues
+  ): Promise<Float32Array> {
+    let modelValues = getModelValues(this.checkpointUrl);
```
jesseengel, line 97: This doesn't make much sense to me, it makes the model a leaky abstraction (relying on the values in CONSTANTS). 

It seems like what we would want is to set all of these values in the constructor.

music/src/ddsp/ddsp.ts
```diff
@@ -0,0 +1,387 @@
...
+
+export function getModelValues(checkpointUrl: string): ModelValues {
+  const presetModelValues = PRESET_MODELS.filter(
+    (modelValues: ModelValues): boolean =>
+      modelValues.checkpointUrl === checkpointUrl
+  );
+  return presetModelValues.length > 0 ? presetModelValues[0] : {};
+}
+
+async function synthesize(
```
jesseengel, line 347: Since this function only is used in the context of a model, wouldn't it make sense to just have it as a method of the Model class, rather than a separate function?

music/src/ddsp/ddsp.ts
```diff
@@ -0,0 +1,387 @@
...
+        resizedAudioFeatures.f0_hz[counter] = audioFeatures.f0_hz[i];
+      }
+      counter += 1;
+    }
+
+    resolve(resizedAudioFeatures);
+  });
+}
+
+async function ddsp_process(
```
jesseengel, line 107: Same for this one, it only really makes sense in the context of a model so it seems cleaner to just make it a method of the model class.

music/src/ddsp/ddsp.ts
```diff
@@ -0,0 +1,387 @@
...
+    finalf0.dispose();
+    maskedPowerThreshold.dispose();
+    mask.dispose();
+    confMask.dispose();
+  }
+
+  return { f0_hz, loudness_db };
+}
+
+export function getModelValues(checkpointUrl: string): ModelValues {
```
jesseengel, line 339: This is part of that leaky abstraction, and should probably take place in the model class constructor.

music/src/ddsp/constants.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export const MODEL_URL =
+  'https://storage.googleapis.com/magentadata/js/checkpoints/ddsp';
+
+export const LOWEST_LD = -120;
+export const CROSSFADE_DURATION = 1;
+export const OUTPUT_SAMPLE_RATE = 48000;
+
+export const PRESET_MODELS = [
```
jesseengel, line 34: More about the leaky abstraction / constructor. Taking a new look at this, I don't think having these as constants in the javascript code makes a lot of sense.

We'd like the ability to let people train their own models soon, and this would require them to add these values to the magenta-js code which doesn't work well for dropping in new models at runtime. 

I think it might make sense to have these in a json object (stored on cloud) that is read as well as the model weights. That way we can have our training scripts create these extra json files as well, and then users can just point to the URL of the directory with all the model info in it. 

I think ideally, the model constructor would take the URL and these values in it's constructor, but then override these values with values from the JSON if it's found.

We could do this as a later PR if you want, but I think it's important to do, and would keep the Model class from being leaky.

jesseengel (commented):
Getting there! Some more comments.

music/src/ddsp/model.ts
EdwinToh: by putting the values here instead of the constructor, we give the users the ability to tweak the values without creating a new instance.

music/src/ddsp/ddsp.ts
EdwinToh: this is a method of the model class, or maybe i am missing your point. the only thing i did here is to separate it out into it's own file to keep the model class a little bit more lightweight. i can move it all into the same file.

music/src/ddsp/constants.ts
EdwinToh: i see what you mean now with the constructor, perhaps we can have it in both places. these options will be read in the constructor and can be overwritten in the synthesize method.

i agree with taking it out of the constants file and moving them somewhere else.   

now that i take a look at this again, i realize that there are no default values if the checkpointUrl doesn't match any of these preset models. perhaps each one of these values need to be required parameters in the constructor and we can list the preset models' values in documentation.

music/src/ddsp/model.ts
jesseengel: Understood, I think the better way to do that is to have the method take in optional arguments on call, that use the model attributes as defaults if not provided.

music/src/ddsp/ddsp.ts
jesseengel: yah, that's it. It just makes sense to have all the code together, because these are all the things specific to the model. (i.e. `model` is always `self`)

music/src/ddsp/constants.ts
jesseengel: yah, I think it makes sense to override with values read them from jsons (that we can create for new datasets and new models as well).

i.e. Order of overwriting:
====================
--> Constructor Defaults
--> Constructor Args
---> Values loaded from json read in constructor

EdwinToh: I've moved these values out of the constants and created a new README to provide these values since we don't have a way to load them from a json alongside the model for now.

EdwinToh:
@jesseengel I've moved the constants out and added a README for DDSP where I provide the values for the 4 respective models. Let me know what you think?

music/src/ddsp/model.ts
```diff
@@ -0,0 +1,243 @@
...
+      const f0hzTensor = tf.tensor1d(f0_hz, 'float32');
+      // tslint:disable-next-line: variable-name
+      const loudnessTensor = tf.tensor1d(loudness_db, 'float32');
+
+      const result = await this.model.predict({
+        f0_hz: f0hzTensor,
+        loudness_db: loudnessTensor,
+      });
+
+      // apply a loudness modifier here to make some
```
jesseengel, line 190: wouldn't it make sense to do this once on the stitched audio instead of on each chunk?

music/src/spice/spice.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+import * as tf from '@tensorflow/tfjs';
+import { resampleAndMakeMono } from '../core/audio_utils';
+import { computePower } from './loudness_utils';
+import { AudioFeatures } from '../ddsp/interfaces';
+import { getPitches } from './pitch_utils';
+
+// SPICE SPECIFIC CONSTANTS
+const SPICE_MODEL_URL = 'https://tfhub.dev/google/tfjs-model/spice/2/default/1';
```
jesseengel, line 25: I like that you made a spice model class. Wouldn't it make sense to have these be default args of the Spice model, rather than in this separate file? and to have the functions be methods of the model? That way it has the same api as the other models (takes in a url in the constructor, but with a default to this url).

music/src/ddsp/README.md
```diff
@@ -0,0 +1,85 @@
...
+  }
+);
+
+await violinDDSP.initialize();
+```
+
+To start synthesizing, you'll need to pass in AudioFeatures (derived from the [SPICE](../spice) class).
+
+```
+violinDDSP.synthesize(audioFeatures);
```
jesseengel, line 74: can you also give an example of passing in a different say postGain or meanPitch?

music/src/ddsp/README.md
```diff
@@ -0,0 +1,85 @@
+# DDSP
+
+## Preset Checkpoints
+
+There are [4 DDSP Preset models](../../checkpoints) you can use, you can also create and use your own.
+
+Each of these models need some settings to be passed. You can tweak these settings, but these are the defaults:
+
+```
+  // violin
```
jesseengel, line 10: Okay I've added all of these settings as `settings.json` files in each corresponding folder. I like that you provide documentation here on how to build with your own settings in the constructor, but I think the documentation and code should default to getting the URL of the folder (like in musicvae: https://github.com/magenta/magenta-js/blob/master/music/demos/music_vae.ts#L32), and then reading $URL/model.json and $URL/settings.json from the folder (sort of like in musicvae: https://github.com/magenta/magenta-js/blob/master/music/src/music_vae/model.ts#L847). The model should only error out if it didn't receive a settings in either the constructor or find and parse a settings.json in folder.

music/src/ddsp/README.md
```diff
@@ -0,0 +1,85 @@
...
+    modelMaxFrameLength: 1250,
+  }
+```
+
+## Using DDSP
+
+To start using DDSP, create a new instance of DDSP and initialize it:
+
+```
+const violinDDSP = new DDSP(
```
jesseengel, line 56: It's cool to show an example of manually setting it, but it should be the second example, where the first is just

```
const violinDDSP = new DDSP(
  "https://storage.googleapis.com/magentadata/js/checkpoints/ddsp/violin"
);
await violinDDSP.initialize();
```

jesseengel (commented):
Okay, getting real close! I added `settings.json` files to cloud and left comments

music/src/ddsp/model.ts
EdwinToh: yes, good point! however, looking at the code, it seems like we're doing this mul on a tensor and then disposing it once we get the data. 

if we do it once at the end, we'll have to convert audioChunks into a tensor, run the mul, then take the data again.  not sure if there is a performance gain if we do that.

music/src/spice/spice.ts
EdwinToh: I think it makes more sense to have this here since this constant won't be modified (and shouldn't be). I think to try to make this class be consistent with the other classes in the Magenta library is a bad idea in this case because the url parameter will be optional and might potentially mislead one into thinking that they can use their own model somehow. 

I'm in favor to keep it this way unless you feel strongly about this.

music/src/ddsp/README.md
EdwinToh: Nice! I've modified the class to do this now.

music/src/spice/spice.ts
EdwinToh: on second thought, users might want to avoid using tfhub altogether. i've made the changes as you suggested!

music/src/ddsp/model.ts
jesseengel: okay, I was just looking for code clarity mostly. If you think it will look/be worse, no worries.

EdwinToh: word. i've made the change. I realized that it was a simple multiplier and doesnt need a tensor to do this operation.

EdwinToh:
@jesseengel @notwaldorf Looks like everything is addressed. Let me know if there are anymore things to take a look at!

music/src/core/audio_utils.ts
```diff
@@ -453,7 +504,7 @@ function outerSubtract(arr: Float32Array, arr2: Float32Array): Float32Array[] {
 }
 
 function pow(arr: Float32Array, power: number) {
-  return arr.map(v => Math.pow(v, power));
+  return arr.map((v) => Math.pow(v, power));
```
notwaldorf, line 507: What code formatter are you using? Is it the project default one (clang-format)? there's a lot of these meaningless formatting changes that I'm not sure where they're coming from...

music/.gitignore
```diff
@@ -5,3 +5,5 @@ es6/
 node/
 dist/
 yarn-error.log
+report.*.json
+tsconfig.json
```
notwaldorf, line 9: why is the tsconfig being ignored? seems bad

music/src/core/logging.ts
```diff
@@ -15,23 +15,34 @@
...
 
 /**
  * The different verbosity levels.
  */
 export const enum Level {
-  NONE = 0,   // No messages will be logged.
-  WARN = 5,   // WARN messages will be logged.
-  INFO = 10,  // INFO and WARN messages will be logged.
-  DEBUG = 20  // DEBUG, INFO, and WARN messages will be logged.
+  NONE = 0, // No messages will be logged.
```
notwaldorf, line 24: This code formatting change seems strictly worse, for example

music/src/core/player.ts
```diff
@@ -409,29 +399,35 @@ class DrumKit {
...
 
-  private bassSynth =
-      new Tone.Synth({volume: 5, oscillator: {type: 'triangle'}}).toMaster();
+  private bassSynth = new Tone.Synth({
+    volume: 5,
+    oscillator: { type: 'triangle' },
+  }).toDestination();
 
-  private polySynth = new Tone.PolySynth(10).toMaster();
+  private polySynth = new Tone.PolySynth().toDestination();
```
notwaldorf, line 407: why is the (10) argument not needed anymore?

music/src/core/player.ts
```diff
@@ -409,29 +399,35 @@ class DrumKit {
...
-        undefined;
+    const velocity = note.hasOwnProperty('velocity')
+      ? note.velocity / constants.MAX_MIDI_VELOCITY
+      : undefined;
 
     if (note.isDrum) {
       this.drumKit.playNote(note.pitch, time, velocity);
     } else {
-      const freq = new Tone.Frequency(note.pitch, 'midi');
+      const freq = Tone.Frequency(note.pitch, 'midi').toFrequency();
```
notwaldorf, line 423: Why is the 'new' not needed anymore?

music/src/core/player.ts
```diff
@@ -522,9 +522,12 @@ export class SoundFontPlayer extends BasePlayer {
     const min = isDrum ? constants.MIN_DRUM_PITCH : constants.MIN_PIANO_PITCH;
     const max = isDrum ? constants.MAX_DRUM_PITCH : constants.MAX_PIANO_PITCH;
     for (let i = min; i <= max; i++) {
-      for (let j = constants.MIN_MIDI_VELOCITY; j < constants.MAX_MIDI_VELOCITY;
-           j++) {
-        ns.notes.push({pitch: i, velocity: j, program, isDrum});
+      for (
+        let j = constants.MIN_MIDI_VELOCITY;
```
notwaldorf, line 526: This is also a pretty dubious formatting change

music/webpack/es5.base.config.ts
```diff
@@ -6,7 +6,7 @@ export const baseConfig = {
     rules: [{
       test: /\.ts$/,
       exclude: /node_modules/,
-      use: {loader: 'ts-loader', options: {configFile: 'tsconfig.es5.json'}}
+      use: {loader: 'ts-loader', options: {configFile: 'tsconfig.es6.json'}}
```
notwaldorf, line 9: I don't think this is correct. This is the es5 config file, and should be using the es5 rules, no?

notwaldorf (changes requested):
I've only really reviewed the non-ddsp changes as a library owner. It's super hard to follow this PR because it touches 37 files, a lot of them which only have code formatting changes, and most that don't seem to come from the google formatter (the default one). Do you maybe have prettier set up or something?

Please let me know if you can't get clang-format to work (should work if you open the magenta-js/music folder in vs code...)

music/src/spice/model.ts
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+      this.modelUrl = TFHUB_SPICE_MODEL_URL;
+    }
+  }
+
+  /**
+   * Loads variables from the checkpoint and builds the model graph.
+   */
+  async initialize() {
+    this.spiceModel = await startSpice(this.modelUrl);
```
notwaldorf, line 50: shouldn't this set this.initialized?

music/src/ddsp/model.ts
```diff
@@ -0,0 +1,263 @@
...
+          'Passing settings is required if you do not have a settings.json file.'
+        );
+      }
+    }
+    this.settings = {
+      ...settings,
+      ...this.settings,
+    };
+    console.log('settings', settings);
+    console.log('this.settings', this.settings);
```
notwaldorf, line 94: shouldn't this function set this.initialized = true?

notwaldorf (changes requested):
more comments

music/.gitignore
EdwinToh: afaik, vscode only takes tsconfig.json as the config but since this code repository only has the es5 and 6 flavors, i didn't want to include another tsconfig.json for myself just to see errors inline in my editor.

https://code.visualstudio.com/docs/typescript/typescript-compiling#_tsconfigjson

music/src/core/logging.ts
EdwinToh: i do have prettier setup and personally have always preferred the consistent formatting that prettier gives me and my team. i wasn't aware that the repo itself has another formatter since it never kicked in for me even when i disabled prettier. let me see whether i can setup clang and resave these files.

music/src/core/player.ts
EdwinToh: it's a new tonejs version change: `PolySynth does not require a polyphony value. Voice allocation and disposing is done automatically based on demand.`

music/src/core/player.ts
EdwinToh: because this is now a method and not a class

music/.gitignore
EdwinToh: prob shd also link to the issue: https://github.com/Microsoft/vscode/issues/12463

music/webpack/es5.base.config.ts
EdwinToh: upgrading tonejs (which now is in typescript) started to give errors and some googling gave me this result:
https://stackoverflow.com/questions/50203369/class-constructor-cannot-be-invoked-without-new-typescript-with-commonjs/50203532#50203532 specifically this: ```This problem should be solved in Node.js by setting TypeScript target option to es6 or higher. Modern Node.js versions support ES6 classes, there is no need to transpile them.
```

Happy to change this back to a working es5 tsconfig but I'll require some help from you to get it right.

File: music/src/spice/model.ts (hunk 41 above)
Comment (EdwinToh) [reply music/src/spice/model.ts:50]: good catch!

File: music/src/ddsp/model.ts (hunk 42 above)
Comment (EdwinToh) [reply music/src/ddsp/model.ts:94]: good catch!

File: music/src/core/audio_utils.ts (hunk 34 above)
Comment (EdwinToh) [reply music/src/core/audio_utils.ts:479]: I've re-saved all the files I touched with the clang-formatter and if you still see a lot of changes, it's because of the tonejs and tfjs upgrade. 

The change that touched the most file was the deprecation of `toMaster`, which required me to change it to `toDestination`.

File: music/checkpoints/checkpoints.json [hunk 43]
```diff
@@ -215,5 +215,33 @@
     "description": "A unidirectional piano transcription model. Quantized to 2-byte weights.",
     "url": "https://storage.googleapis.com/magentadata/js/checkpoints/transcription/onsets_frames_uni_q2",
     "size_mb": 30
+  },
+  {
+    "id": "ddsp_flute",
+    "model": "DDSPFlute",
+    "description": "A flute model for use with DDSP.",
+    "url": "https://storage.googleapis.com/magentadata/js/checkpoints/ddsp/flute/model.json",
```
jesseengel, line 223: I think these URLs should be just to the folder right? Not to the model.json?

music/demos/ddsp_tone_transfer.ts
```diff
@@ -0,0 +1,253 @@
...
+  TRUMPET = 'trumpet',
+  FLUTE = 'flute',
+}
+
+export const PRESET_MODEL_URL =
+  'https://storage.googleapis.com/magentadata/js/checkpoints/ddsp';
+
+export const PRESET_MODELS = {
+  [MODEL.VIOLIN]: {
+    checkpointUrl: `${PRESET_MODEL_URL}/${MODEL.VIOLIN}`,
```
jesseengel, line 34: If you have the URL, you shouldn't need these settings, because you can just read them from settings.json off of cloud in the model directory, right?

jesseengel (commented):
Cool, looking pretty good! Just a couple small comments on my side

EdwinToh:
@jesseengel good catches. I've addressed your last 2 comments. 

@notwaldorf Thanks for your reviews so far. it'll be great if you can take another look again when you're free!

music/demos/ddsp_tone_transfer.ts
```diff
@@ -0,0 +1,201 @@
...
+  VIOLIN = 'violin',
+  TENOR_SAXOPHONE = 'tenor_saxophone',
+  TRUMPET = 'trumpet',
+  FLUTE = 'flute',
+}
+
+export const PRESET_MODEL_URL =
+    'https://storage.googleapis.com/magentadata/js/checkpoints/ddsp';
+
+export const PRESET_MODELS = {
```
notwaldorf, line 32: I don't think you really need this separate array. Since you use it as `PRESET_MODELS[MODEL.VIOLIN]`, that value is exactly just `${PRESET_MODEL_URL}/${MODEL.TRUMPET}`. Like, you're not getting anything from the array itself.

music/src/core/audio_utils.ts
notwaldorf: Thanks! Looks good now!

music/src/core/player.ts
notwaldorf: Got it, thanks!

music/src/core/player.ts
```diff
@@ -475,9 +480,10 @@ export class SoundFontPlayer extends BasePlayer {
 
   constructor(
       soundFontURL: string, output = Tone.Master,
-      programOutputs?: Map<number, any>,      // tslint:disable-line:no-any
-      drumOutputs?: Map<number, any>,         // tslint:disable-line:no-any
-      callbackObject?: BasePlayerCallback) {  // tslint:disable-line:no-any
+      programOutputs?: Map<number, any>,  // tslint:disable-line:no-any
+      drumOutputs?: Map<number, any>,     // tslint:disable-line:no-any
+      callbackObject?: BasePlayerCallback) {
+    // tslint:disable-line:no-any
```
notwaldorf, line 486: i don't think this last comment is needed?

music/src/ddsp/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+export const MIN_VRAM = 50;  // used for memory check
+
+export const LOWEST_LD = -120;
+export const CROSSFADE_DURATION = 1;
+export const OUTPUT_SAMPLE_RATE = 48000;
+
+// optimal values
```
notwaldorf, line 24: nit: Capitalize comment, add period at the end

music/src/ddsp/ddsp.ts
```diff
@@ -0,0 +1,225 @@
...
+import {CONF_SMOOTH_SIZE, CONF_THRESHOLD, LD_CONF_REDUCTION, LOWEST_LD, MIN_VRAM,} from './constants';
+import {AudioFeatures, ModelValues} from './interfaces';
+
+async function memCheck() {
+  const bytesPerMB = 1024 * 1024;
+  const screenResToVRAMFactor = 600;
+  const screenSize = window.screen.availWidth * window.screen.availHeight;
+  const DPI = window.devicePixelRatio;
+
+  // checking of vram ensures that the device can hold
```
notwaldorf, line 33: nit: Capitalize comment, add period at the end (in whole file)

music/src/ddsp/model.ts
```diff
@@ -0,0 +1,232 @@
...
+      });
+      const resultData = await (result as Tensor).data();
+      audioChunks.push(resultData);
+
+      (result as Tensor).dispose();
+      f0hzTensor.dispose();
+      loudnessTensor.dispose();
+    }
+
+    // does not require stiching
```
notwaldorf, line 194: nit: Capitalize comment, add period at the end (and rest of file)

music/src/spice/index.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
notwaldorf, line 3: 2020

music/src/spice/pitch_utils.ts
```diff
@@ -0,0 +1,182 @@
...
+    pitchedInput.push(buffer[i]);
+    for (let j = 1; j < dupCountPitches; j++) {
+      pitchedInput.push(-1);
+    }
+    if (i < modulos) {
+      pitchedInput.push(-1);
+    }
+  }
+
+  // cover missing pitches
```
notwaldorf, line 49: nit: Capitalize comment, add period at the end

music/webpack/es5.base.config.ts
notwaldorf: Hmm, i don't think the node build should be using the es5.base.config. I am worried that if we're changing it here, then our es5 build will not be an es5 build anymore.

notwaldorf (commented):
Looks great, thanks for your patience! I basically only left nits. I have a question: have you tested this on mobile? We've had a bunch of problems with tfjs on iphones in particular (float32 errors), and I want to make sure we test this while it's still fresh in someone's head

music/demos/ddsp_tone_transfer.ts
EdwinToh: good point. this array had some settings before, but you're right that it doesn't make sense to keep it without those settings now.

music/src/core/player.ts
EdwinToh: oh weird. this is clang doing some weird formatting but yeah it wasn't needed in the first place

EdwinToh:
@notwaldorf Yep, we checked extensively when we released the Tone Transfer showcase. This works on mobile devices. The float32 issue has to do with iPhones only doing float16 I believe, and we've built against that.

music/webpack/es5.base.config.ts
EdwinToh: thanks for catching this. 

i changed this back to the es5 config and for some reason, I'm no longer running into the issues that I was having before! it's very perplexing, so I went through all the demoes to make sure everything still works and... they do. perhaps the issue I was having had to do with something else which I fixed.

notwaldorf (approved):
It's a :shipit: from me!!!

EdwinToh:
Awesome! Thanks @notwaldorf !

@jesseengel Will either one of you merge this PR? I don't have write access. :)

PR #528, opened 2021-01-05
Fix refs to tone

notwaldorf:
Tone.js had a breaking change, magenta.js got updated for it, so we need to tell people to use the latest version that works.


notwaldorf:
Oops not this one

PR #532, opened 2021-01-10
ESM build

cifkao:
Adding an ESModule build.

Tested this with [example-magenta-in-ts](https://github.com/notwaldorf/example-magenta-in-ts/) (importing from `esm` instead of `es6`) and it seems to work. At first it didn't because of an error in `core/compat`. The reason was that `isNode` in `core/compat/global` was true, because `global` was not undefined (maybe it has something to do with the fact that the module itself is called `global`?). I'm also fixing that problem in this PR.


notwaldorf (commented):
Wooo this looks good! thank you so much for figuring it out!

notwaldorf (approved):

PR #533, opened 2021-01-10
Clean up player and visualizer imports

cifkao:
Cleaning up some imports so that `player` and `visualizer` do not depend on TensorFlow.js via other modules.

Combined with #532, this will make it possible to reduce bundle size quite drastically in cases where we don't need TensorFlow (notwaldorf/example-magenta-in-ts#6).

### DIFF
--- a/music/src/core/data.ts
+++ b/music/src/core/data.ts
@@ -28,31 +28,11 @@ import * as tf from '@tensorflow/tfjs';
 import {INoteSequence, NoteSequence} from '../protobuf/index';
 
 import * as constants from './constants';
+import {DEFAULT_DRUM_PITCH_CLASSES} from './constants';
 import * as logging from './logging';
 import {Melody, MelodyControl, MelodyRhythm, MelodyShape} from './melodies';
 import * as performance from './performance';
 import * as sequences from './sequences';
-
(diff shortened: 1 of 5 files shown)


music/src/core/data.ts
```diff
@@ -28,32 +28,12 @@ import * as tf from '@tensorflow/tfjs';
 import {INoteSequence, NoteSequence} from '../protobuf/index';
 
 import * as constants from './constants';
+import {DEFAULT_DRUM_PITCH_CLASSES} from './constants';
 import * as logging from './logging';
 import {Melody, MelodyControl, MelodyRhythm, MelodyShape} from './melodies';
 import * as performance from './performance';
 import * as sequences from './sequences';
 
-export const DEFAULT_DRUM_PITCH_CLASSES: number[][] = [
```
notwaldorf, line 36: ughhhhhhhh I think this is technically a breaking change. Can we still re-export them from here too, so that all the things that import the constants from here don't break? Sorry, this is the curse of maintaining open source 😭

notwaldorf (changes requested):
This looks good except for the one tiny comment about not breaking everyone using the current export.

notwaldorf (approved):

PR #534, opened 2021-01-12
Add control change data in `midiToSequenceProto`

e7mac:
Currently, control change data is skipped while parsing MIDI data to proto. This PR adds that ability.


e7mac:
@googlebot I signed it!

PR #685, opened 2024-01-10
Export soundfont

OndrejSpanel:
Make soundfond functionality accessible as `core.soundfont` (or `mm.soundfont`). This is useful when someone wants to create a player similiar to `SoundFontPlayer`.

Current workaround is to use the `SoundFontPlayer`, let it create the `SoundFont` instance and then use it in the derived class via the `soundfont` member.


OndrejSpanel:
CLA signed

